PR #5, opened 2019-08-01
Boilerplate changes

ripecosta:
- Inverted the order of the two commands ran by prepublishOnly
- Added myself to contributors
- Added eslint to vscode suggested plugins
- Added `@typescript-eslint/prefer-interface` lint rule

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,6 +21,7 @@
         "prettier/prettier": 2,
         "import/no-default-export": 2,
         "interface-over-type-literal": 0,
+        "@typescript-eslint/prefer-interface": 0,
         "import-name": 0,
         "object-literal-sort-keys": 0,
         "interface-name": 0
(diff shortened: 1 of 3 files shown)


.eslintrc.json
```diff
@@ -21,6 +21,7 @@
         "prettier/prettier": 2,
         "import/no-default-export": 2,
         "interface-over-type-literal": 0,
+        "@typescript-eslint/prefer-interface": 0,
```
GHaberis: 🤔 `0` is equal to turning the rule off, if we want interfaces over type literals should we be turning `"interface-over-type-literal"` and `"@typescript-eslint/prefer-interface"`, to 2? 2 throws an error if the rule is broken and 1 shows a warning.

ripecosta: I wanted the linter not to prefer using interface instead of type and allow the developer to use whatever they prefer, I turned it off. Do you disagree with this approach?

GHaberis: No i'm happy to go with this, we haven't discussed whether we'd use `interface` over `type` literals so I agree we shouldn't be enforcing anything for now.

PR #9, opened 2019-08-27
Use js-cookie library

GHaberis:
This PR introdices the js-cookie library. This provides a simple API for saving and retrieving cookies, it is lightweight (g-zipped ~1.1k) and it ensures the values saved are [RFC-6265](https://tools.ietf.org/html/rfc6265#section-4.1.1) compliant.

### DIFF
--- a/src/cookies.ts
+++ b/src/cookies.ts
@@ -1,6 +1,4 @@
-// subset of https://github.com/guzzle/guzzle/pull/1131
-const isValidCookieValue = (name: string): boolean =>
-    !/[()<>@,;"\\/[\]?={} \t]/g.test(name);
+import * as Cookies from 'js-cookie';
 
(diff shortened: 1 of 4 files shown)


src/cookies.ts
```diff
@@ -15,56 +13,25 @@ const getShortDomain = (): string => {
...
-    const expires = new Date();
-
-    if (!isValidCookieValue(name) || !isValidCookieValue(value)) {
-        return;
-    }
+    const options: {
+        domain: string;
+        expires?: number;
+    } = {
+        domain: getDomainAttribute(),
```
ripecosta: What happens when we give it `domain: ''`? Does it act the same way as if no domain option has been given?

GHaberis: The library doesn't add the `domain` attribute if it  is an empty string

PR #18, opened 2019-09-03
Update IAB Vendor List link

ripecosta:
...to the correct Frontend endpoint, that updates automatically.


src/config.ts
```diff
@@ -19,7 +19,7 @@ export const COOKIE_MAX_AGE = 395; // 13 months
 export const GU_AD_CONSENT_COOKIE = 'GU_TK';
 export const GU_COOKIE_NAME = 'guconsent';
 export const IAB_VENDOR_LIST_URL =
-    'https://assets.guim.co.uk/data/vendor/4f4a6324c7fe376c17ceb2288a84a076/cmp_vendorlist.json';
+    'https://github.com/guardian/manage-frontend/pull/263';
```
adamnfish: Pretty sure this is wrong!

ripecosta: Fixed!

PR #35, opened 2019-09-23
onIabConsentNotification return IAB state derived from IAB Cookie if available

GHaberis:
This PR updates the `cmp.ts` module. 

Callbacks currently passed to `onIabConsentNotification` are executed with the `iabPurposeRegister.state` as their first argument. Currently the value of `iabPurposeRegister.state` is derived from the `GU_TK` cookie. For example, If `GU_TK === '1.54321'` then `iabPurposeRegister.state = {1: true, 2: true, 3: true, ...}` OR If `GU_TK === '0.54321'` then `iabPurposeRegister.state = {1: false, 2: false, 3: false, ...}`.

This PR updates the `cmp.ts` module so callbacks passed to `onIabConsentNotification` are now executed with an `iabPurposeRegister.state` derived from the IAB string cookie if it exists. If the IAB string cookie does not exist then we then fallback to trying the `GU_TK` cookie (as before).

### DIFF
--- a/src/cmp.test.ts
+++ b/src/cmp.test.ts
@@ -50,20 +50,12 @@ describe('cmp', () => {
     });
 
     describe('onIabConsentNotification', () => {
-        beforeEach(() => {
-            ConsentString.mockImplementation(() => {
-                return {
-                    isPurposeAllowed: jest.fn(() => true),
-                };
-            });
-        });
-
         describe('if no IAB cookie available', () => {
             beforeEach(() => {
                 readIabCookie.mockReturnValue(null);
             });
 
-            it('executes advertisement callback with true if GU_TK cookie is true', () => {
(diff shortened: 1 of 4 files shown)


src/cmp.test.ts
```diff
@@ -96,24 +103,133 @@ describe('cmp', () => {
...
-                Cookies.get.mockReturnValue('1.54321');
+            it('executes advertisement callback each time consent nofication triggered with updated state', () => {
+                Cookies.get
+                    .mockReturnValue('1.54321') // default
+                    .mockReturnValueOnce(undefined); // first call
 
                 const myCallBack = jest.fn();
                 const expectedArguments = [
                     [
+                        {
```
ripecosta: We use these 5 null, 5 true or 5 false for the IAB objects a lot. Should we create 3 objects with that easy to so it's easier to read?

PR #37, opened 2019-10-07
Add functionality to Post to Consent Logs

GHaberis:
This PR updates the CMP library so it is responsible for posting consent data to the consent logs and also saving the cookie. Previously the CMP UI had responsibility for these two tasks however we found that posts to the consent logs were being cancelled as the CMP UI was removed from the DOM by it's parent (in this case, frontend) before the associated network request had a response.

By moving responsibility for posting consent data to the library these posts will now originate  from frontend meaning we can remove the CMP UI without the network requests being cancelled. We decided to move storing the Cookie to the library too as the data to be stored was now being sent from the CMP UI to it's parent so was available to the library running on the parent.

To achieve this we update the format of postMessages originating from the CMP UI, from `string` to objects: 

```js
{
msgType: CMP_READY_MSG,
msgData: { .. }
}
```

in the msgData for a `CMP_READY_MSG` we now pass the data required when saving the cookie and posting to the consent logs. 

We've also included a 3rd argument to `setupMessageHandlers` in `cmp-ui` called `onErrorCmp`, this a callback that is executed when an Error is thrown, it allows users of the library to catch errors and handle them with their own custom reporting tools, eg. `Raven` in `frontend`.


src/cmp-ui.test.ts
```diff
@@ -1,15 +1,15 @@
 import { setupMessageHandlers, canShow } from './cmp-ui';
 import { CMP_DOMAIN, CMP_READY_MSG, CMP_CLOSE_MSG } from './config';
 import {
-    readGuCookie as _readGuCookie,
+    // readGuCookie as _readGuCookie,
```
adamnfish: Do these need to be commented out here, or is it better to just remove them>

src/cmp-ui.test.ts
```diff
@@ -111,27 +119,36 @@ describe('cmp-ui', () => {
...
             expect(canShow()).toBe(true);
         });
 
-        it('canShow returns false if readIabCookie and readGuCookie return truthy values', () => {
+        it('canShow returns false if readIabCookie returns truthy value', () => {
             readIabCookie.mockReturnValue('foo');
-            readGuCookie.mockReturnValue('bar');
             expect(canShow()).toBe(false);
         });
+        // it('canShow returns true if readGuCookie returns null', () => {
```
adamnfish: Are we missing these test cases now? As above, if they are redundant do they need to be here commented out?

src/cmp-ui.ts
```diff
@@ -1,25 +1,55 @@
...
-type Callback = () => void;
+type Callback = (error?: Error) => void;
 
 export const setupMessageHandlers = (
     onReadyCmp: Callback,
     onCloseCmp: Callback,
+    onErrorCmp: Callback,
 ): void => {
     const receiveMessage = (event: MessageEvent): void => {
+        const withError = (callback: Callback): void => {
```
adamnfish: This might be better named as something like "with errors handled" or even "without errors", given its functionality?

src/cmp-ui.ts
```diff
@@ -1,25 +1,55 @@
...
                 break;
             case CMP_CLOSE_MSG:
-                onCloseCmp();
+                withError(onCloseCmp);
+                break;
+            case CMP_SAVED_MSG:
+                logConsent(msgData)
+                    .then(response => {
+                        if (!response.ok) {
+                            throw new Error(
```
adamnfish: Is this error handled (e.g. logged) somewhere else? Be good to have visibility of these if possible.

src/cmp-ui.ts
```diff
@@ -29,4 +59,4 @@ export const setupMessageHandlers = (
     window.addEventListener('message', receiveMessage, false);
 };
 
-export const canShow = (): boolean => !readGuCookie() || !readIabCookie();
+export const canShow = (): boolean => !readIabCookie(); // TODO: Restore readGuCookie check once we start saving GU cookie
```
adamnfish: What does this TODO mean?

src/cmp.test.ts
```diff
@@ -106,10 +106,38 @@ describe('cmp', () => {
                 expect(myCallBack).toBeCalledWith(iabNullState);
             });
 
+            //     it('executes advertisement callback each time consent nofication triggered with updated state', () => {
```
adamnfish: Seems like this test case was testing an important piece of behaviour. Is this now covered elsewhere?

src/consent-logs.ts
```diff
@@ -0,0 +1,84 @@
...
+    const browserID = Cookies.get('bwid') || DUMMY_BROWSER_ID;
+
+    if (isProd && browserID === DUMMY_BROWSER_ID) {
+        throw new Error(`Error getting browserID in PROD`);
+    }
+
+    const logInfo = {
+        version: '1',
+        iab: consentStr,
+        source: 'www',
```
adamnfish: This is fine for now, but this will definitely need updating before it can be rolled out more widely. It might be good to set the interface to this function now (and pass in `www` in the only case we have), rather than it be hard coded in the lib so we have to line up a change at both ends in the near future.

Up to you though, no worries at all if you're happy to go out like this and refactor later.

src/consent-logs.ts
```diff
@@ -0,0 +1,84 @@
...
+        purposes: {
+            personalisedAdvertising: pAdvertising,
+        },
+        browserId: browserID,
+        variant: 'CmpUiIab-variant',
```
adamnfish: This might be better coming from config if it's going to be hard coded at this end. Not important for now, but in the future we should think about the use cases of testing the banner and work out whether this should be a hard-coded here or provided by a test framework.

src/cmp-ui.test.ts
GHaberis: I commented these out as the exclusion of `readGuCookie()` from the `canShow` check in `cmp-ui.ts` is temporary, whilst we're not handling consent information for the PECR purposes. I felt leaving them here but commented out gave us an indication of what we intend to be doing once we start handling consent information for the PECR purposes.

src/cmp-ui.test.ts
GHaberis: Linked to my comment https://github.com/guardian/consent-management-platform/pull/37/files#r332073439 these tests will be useful once we restore `readGuCookie()` to the `canShow` check.

src/cmp-ui.ts
GHaberis: This error will be caught by the `catch` block which will pass it onto `onErrorCmp`, this function is passed into the library from `frontend` and logs these errors in Sentry.

src/cmp-ui.ts
GHaberis: As mentioned previously i've left this in as a marker that once we start saving the "guCookie" with the PECR purposes we'll want to include `readGuCookie()` in the `canShow` check.

src/cmp.test.ts
GHaberis: Good spot, this test has been duplicated and updated, the commented out version here can be removed

src/consent-logs.ts
GHaberis: This was only ever going to be hardcoded for the duration of the first test. Again `variant` will eventually come from elsewhere, a decision needs to be made as to whether it'll come back from the CMP UI in the `postMessage`, or whether it'll be set some other way. For now I'm happy to leave it here and refactor out once the test is out.

PR #40, opened 2019-11-04
Pull the temporary legacy cookie reading code into cookie.ts

GHaberis:
Pulls the temporary legacy cookie reading code into `cookie.ts` by exposing new `readLegacyCookie` function on it.

### DIFF
--- a/src/cookies.test.ts
+++ b/src/cookies.test.ts
@@ -2,17 +2,22 @@ import * as _Cookies from 'js-cookie';
 import {
     readGuCookie,
     readIabCookie,
+    readLegacyCookie,
     writeGuCookie,
     writeIabCookie,
+    writeLegacyCookie,
 } from './cookies';
 import {
     GU_COOKIE_NAME,
     GU_COOKIE_VERSION,
     IAB_COOKIE_NAME,
     COOKIE_MAX_AGE,
+    LEGACY_COOKIE_NAME,
 } from './config';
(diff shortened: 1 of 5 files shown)


src/cookies.test.ts
```diff
@@ -2,18 +2,23 @@ import * as _Cookies from 'js-cookie';
...
     GU_COOKIE_NAME,
     GU_COOKIE_VERSION,
     IAB_COOKIE_NAME,
     COOKIE_MAX_AGE,
+    LEGACY_COOKIE_NAME,
 } from './config';
 
 const Cookies = _Cookies;
 
+const OriginalDate = global.Date; // Mock the Date constructor to always return the beginning of time
```
ripecosta: Is this comment correct? And do we still need to do this since we're mocking Date.now() only?

src/cookies.test.ts
```diff
@@ -36,11 +41,21 @@ describe('Cookies', () => {
...
+            now: () => '12345',
+        };
         Object.defineProperty(document, 'domain', {
             value: 'www.theguardian.com',
         });
     });
 
+    afterAll(() => {
+        global.Date = OriginalDate;
+        expect(new Date().toString()).not.toMatch(
```
ripecosta: This test doesn't make sense to me. We are not setting Date() to the beginning of time so it will always return true.

PR #42, opened 2019-11-05
CMP restructure Phase 2

ripecosta:
This PR executes phase two of the CMP restructure, which includes:
- Rename `cmp.ts` to `core.ts`
- Move the consent state logic into `store.ts`
- Add unit testing for both files

Notes:
- As things move places the code in `consent-storage.ts` that triggered the consent state notification was commented. This module will be refactored in Phase 3.


src/store.test.ts
```diff
@@ -0,0 +1,117 @@
...
+        it('when the GU_TK cookie is available', () => {
+            readLegacyCookie.mockImplementation(() => legacyTrueCookie);
+
+            expect(getConsentState()).toMatchObject({
+                guState: guDefaultState,
+                iabState: iabTrueState,
+            });
+        });
+        it('after a state change', () => {
+            setConsentState(guMixedState, iabMixedState);
```
GHaberis: You could put an expectation checking that `getConsentState()` returns the default states before this `setConsentState` call, this would demonstrate the state has updated as expected when the second `getConsentState()` call occurs.

PR #45, opened 2019-11-11
CMP restructure phase 5

ripecosta:
This PR executes phase five of the CMP restructure, which includes:

- Pass variant into the UI component, save it in `store.ts` and send it in the logs
- Pass source into the UI component, save it in `store.ts` and send it in the logs
- Expose a `getVariant()` function in `store.ts` 
- Add unit tests for the new functions

Notes:
- Moved some code around inside `store.ts` to improve its organisation, such as grouping similar variables together and grouping setter and getter functions.
- Fixed an issue in `logs.ts` that would send an empty string in the variant when the value passed is null or undefined. It now removes the field from the payload entirely, as expected by the consent logs API.

### DIFF
--- a/src/component/ConsentManagementPlatform.tsx
+++ b/src/component/ConsentManagementPlatform.tsx
@@ -4,7 +4,7 @@ import { mobileLandscape, palette, space } from '@guardian/src-foundations';
 import { Logo } from './svgs/Logo';
 import { ConsentPreferencesDashboard } from './ConsentPreferencesDashboard';
 import { SCROLLABLE_ID, CONTENT_ID } from './utils/config';
-import { setSource } from '../store';
+import { setSource, setVariant } from '../store';
 
 const TRANSITION_TIME = 1000;
 
@@ -150,7 +150,7 @@ class ConsentManagementPlatform extends Component<Props, State> {
         }
 
         if (props.variant) {
-            setSource(props.variant);
+            setVariant(props.variant);
         }
     }
 
--- a/src/component/ConsentManagementPlatform.tsx
+++ b/src/component/ConsentManagementPlatform.tsx
@@ -4,6 +4,7 @@ import { mobileLandscape, palette, space } from '@guardian/src-foundations';
 import { Logo } from './svgs/Logo';
 import { ConsentPreferencesDashboard } from './ConsentPreferencesDashboard';
 import { SCROLLABLE_ID, CONTENT_ID } from './utils/config';
+import { setSource } from '../store';
 
 const TRANSITION_TIME = 1000;
 
@@ -127,6 +128,8 @@ interface State {
 
 interface Props {
     onClose: () => void;
+    source?: string;
+    variant?: string;
 }
 
 class ConsentManagementPlatform extends Component<Props, State> {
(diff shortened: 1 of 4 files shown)


src/component/ConsentManagementPlatform.tsx
```diff
@@ -141,6 +144,14 @@ class ConsentManagementPlatform extends Component<Props, State> {
             active: false,
             visible: false,
         };
+
+        if (props.source) {
+            setSource(props.source);
+        }
+
+        if (props.variant) {
+            setSource(props.variant);
```
GHaberis: I think you should be importing and calling `setVariant` here?

src/store.ts
```diff
@@ -22,15 +23,17 @@ const IAB_VENDOR_LIST_PROD_URL =
     'https://www.theguardian.com/commercial/cmp/vendorlist.json';
 const IAB_VENDOR_LIST_NOT_PROD_URL =
     'https://code.dev-theguardian.com/commercial/cmp/vendorlist.json';
+
+let initialised = false;
+let source = DEFAULT_SOURCE;
+let variant: string | null = null;
```
GHaberis: Question, should the default be null? If a user is not under test can we not include the `variant` property in the data we post to the consent logs or do we just set it to an empty string: `variant: ''`? @janua should be able to answer the prefered option?

src/store.ts
```diff
@@ -121,62 +179,35 @@ const setConsentState = (
         });
 };
 
-const getGuStateFromCookie = (): GuPurposeState => {
-    // TODO: Friendly reminder to update the getConsentState() unit tests after PECR is introduced
-    return { functional: true, performance: true };
+const setSource = (newSource: string) => {
```
GHaberis: `setSource` is missing a return type;

src/store.ts
```diff
@@ -121,62 +179,35 @@ const setConsentState = (
...
-        return newIabState;
-    }
-
-    Object.keys(iabState).forEach((key: string): void => {
-        const purposeId = parseInt(key, 10);
-        newIabState[purposeId] = null;
-    });
-
-    return newIabState;
+const setVariant = (newVariant: string) => {
```
GHaberis: `setVariant ` is missing a return type;

src/component/ConsentManagementPlatform.tsx
ripecosta: 👍

src/store.ts
ripecosta: I set it to null as opposed to undefined so it's just as easy to check against and way easier to reset when we're unit testing the module. But what we're sending to the logs when this value is set to null is `{ variant: "" }` (check line 161 of this file).
I was working under the impression that we always need to pass a variant field but not sure if it was something @janua told us or something I assumed. Would be great if he could confirm.

src/store.ts
ripecosta: 👍

src/store.ts
ripecosta: 👍

src/store.ts
janua: @GHaberis @ripecosta The consent logs side is either expecting a non-empty string value or the key completely missing. Null is unfortunately not accepted right now.

ripecosta: 👍 Will make the changes.

PR #48, opened 2019-11-15
CMP restructure phase 6

ripecosta:
This PR executes phase six of the CMP restructure, which includes:
- Add error handling
- Include `setErrorHandler` as a package export
- Add/update unit testing for affected modules

Notes:
- Updated functions that returns a promise to reject said promise on cases where the request failed.
- Improved how we test promise resolving/rejection.

### DIFF
--- a/src/component/ConsentPreferencesDashboard.tsx
+++ b/src/component/ConsentPreferencesDashboard.tsx
@@ -248,12 +248,8 @@ export class ConsentPreferencesDashboard extends Component<Props, State> {
             .then(remoteVendorList => {
                 return this.buildState(parseIabVendorList(remoteVendorList));
             })
-            .then(() => {
-                showCmp();
-            })
-            .catch(() => {
-                hideCmp();
-            });
+            .then(showCmp)
+            .catch(hideCmp);
     }
 
     public render(): React.ReactNode {
--- a/src/component/ConsentPreferencesDashboard.tsx
+++ b/src/component/ConsentPreferencesDashboard.tsx
@@ -242,19 +242,17 @@ export class ConsentPreferencesDashboard extends Component<Props, State> {
     }
 
     public componentDidMount(): void {
+        const { showCmp, hideCmp } = this.props;
+
         getVendorList()
             .then(remoteVendorList => {
                 return this.buildState(parseIabVendorList(remoteVendorList));
             })
             .then(() => {
-                const { showCmp } = this.props;
-
                 showCmp();
             })
-            .catch(error => {
-                // TODO: Handle error by removing the component
-                // eslint-disable-next-line no-console
-                console.log('Error', error);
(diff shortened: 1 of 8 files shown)


src/component/ConsentPreferencesDashboard.tsx
```diff
@@ -252,7 +252,7 @@ export class ConsentPreferencesDashboard extends Component<Props, State> {
                 showCmp();
             })
             .catch(error => {
-                // TODO: ERROR HANDLING
+                // TODO: Handle error by removing the component
                 // eslint-disable-next-line no-console
                 console.log('Error', error);
```
GHaberis: The `<ConsentManagementPlatform>` component (one level up) takes a prop `onClose` which is a function. We execute this when we close the CMP. What do you think to feeding this into this `<ConsentPreferencesDashboard.tsx>` component as a part of a prop eg. `onVendorListError`. We could execute `onVendorListError` within this catch block. eg.

```js
// ConsentManagementPlatform.tsx
<ConsentPreferencesDashboard.tsx onVendorListError{() => {
  this.props.onClose();
}} />
```

src/error.ts
```diff
@@ -0,0 +1,15 @@
+type onErrorFn = (error: string) => void;
+
+let onError: onErrorFn;
```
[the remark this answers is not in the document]
ripecosta: I had the same idea while working on this actually. I'll make the change.

src/component/ConsentPreferencesDashboard.tsx
ripecosta: Yeah sound good. Will make this change also.

ripecosta: Looking at our code it seems like calling `hideCmp()` would achieve the same without adding an extra prop. I've made that change but was wondering what you think about it. Happy to go a different way if you think it's better.

src/error.test.ts
```diff
@@ -8,6 +8,17 @@ describe('Error', () => {
         jest.resetAllMocks();
     });
 
+    it('runs the default error handler correctly if no handler has been set', () => {
+        /* eslint-disable no-console */
+        console.error = jest.fn();
```
GHaberis: Will `console.error` be reset to it's true/original state following this test? If not I'd suggest we follow the pattern we've used else where for global objects/functions:
```js
global.console = {error: jest.fn()};
expect(console.error).toBeCalled();
global.console.mockClear();
delete global.console;
```

src/component/ConsentPreferencesDashboard.tsx
GHaberis: Happy to go with `hideCmp` in this case

src/error.test.ts
ripecosta: 👍

PR #49, opened 2019-11-19
Introduce FontsContext (CMP restructure)

GHaberis:
Currently the `font-family` names we use in the CSS-in-JS for the CMP UI component are hardcoded to be the same as the `font-family` names used on `frontend`. 

Unfortunately, although the font files on `frontend` are up to date the `font-family` names it uses are the out of date legacy `font-family` names. For example...

Legacy headline font-family: `Guardian Egyptian Web, Georgia, serif`
Up to date headline font-family: `GH Guardian Headline, Georgia, serif` 

Other platforms on which the new CMP UI will run (dotcom-rendering for example) use the up to date files AND the up to date `font-family` names too.

This PR updates the `<ConsentManagementPlatform>` component to take a new prop `fontFamilies`. This optional prop looks like:

```js
<ConsentManagementPlatform fontFamilies={{
  headlineSerif: "GH Guardian Headline, Georgia, serif",
  bodySerif: "GuardianTextEgyptian, Georgia, serif",
  bodySans: "GuardianTextSans, Helvetica Neue, Helvetica, Arial, Lucida Grande, sans-serif",
}} />
```

This prop allows consumers of the `<ConsentManagementPlatform>` component to pass in the correct `font-family` names in use for their platform.

In the `<ConsentManagementPlatform>` component I've created a new React [context](https://reactjs.org/docs/context.html)  called `<FontsContext>`. 

> Context provides a way to pass data through the component tree without having to pass props down manually at every level.

I'm using context so we don't have to pass the `fontFamilies` prop down the component tree to every component that needs to use font styles. Below is an example of how a component that needs to use font styles can "consume" the context:

```js
<FontsContext.Consumer>
  {({ headlineSerif }) => (
    <h1 css={css`font-family: ${headlineSerif}`}>Hello World</h1>
  )}
</FontsContext.Consumer>
```

As mentioned earlier the fontFamilies property passed to  `<ConsentManagementPlatform>` is optional, if nothing is passed the context will use the `DEFAULT_FONT_FAMILIES`, which are the latest `font-family` names. 

**In time hopefully all platforms will use the same `font-family` names and we can remove this requirement to support multiple `font-family` names for the same font file!**

### DIFF
--- a/src/component/ConsentManagementPlatform.tsx
+++ b/src/component/ConsentManagementPlatform.tsx
@@ -31,6 +31,13 @@ const overlayStyles = css`
         transition-duration: ${TRANSITION_TIME / 2}ms;
         will-change: background-color;
     }
+    box-sizing: border-box;
+
+    *,
+    *:before,
+    *:after {
+        box-sizing: inherit;
+    }
 `;
 
 const activeOverlayStyles = css`
--- a/src/component/ConsentManagementPlatform.tsx
+++ b/src/component/ConsentManagementPlatform.tsx
@@ -3,8 +3,14 @@ import { css } from '@emotion/core';
 import { mobileLandscape, palette, space } from '@guardian/src-foundations';
 import { Logo } from './svgs/Logo';
 import { ConsentPreferencesDashboard } from './ConsentPreferencesDashboard';
-import { SCROLLABLE_ID, CONTENT_ID } from './utils/config';
+import { FontsContext } from './FontsContext';
+import {
+    SCROLLABLE_ID,
+    CONTENT_ID,
+    DEFAULT_FONT_FAMILIES,
+} from './utils/config';
 import { setSource, setVariant } from '../store';
+import { FontsContextInterface } from '../types';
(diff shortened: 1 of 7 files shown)


src/component/ConsentManagementPlatform.tsx
```diff
@@ -130,8 +143,8 @@ interface Props {
     onClose: () => void;
     source?: string;
     variant?: string;
+    fontFamilies?: FontsContextInterface;
```
ripecosta: These 3 items are optional but some will be used more often than others so I'd suggest ordering by most likely to be used to least likely to be used. So it would be fontFamilies, source, variant. What do you think?

GHaberis: I'm going to leave this for now as the ordering doesn't make a difference to the execution and the frequency of use is open to change.

ripecosta: I'm okay with that. 👍

PR #54, opened 2019-11-29
Unified build process

ripecosta:
## Current state
We currently have 2 separate build processes:
- One to build the CMP system that uses tsc
- One to build the CMP UI that uses rollup to bundle.

## The problem
Besides being bad practice to have two separate build processes this is causing unwanted effects where 2 different instances of `store.ts` are created as opposed to only one that is shared between the core and the ui parts of the library.

## The goal
We want to avoid importing the UI components unless the UI needs to be shown, which means having 2 bundles (one for the system and one for UI). These two bundles should use the same instance of the `store.ts` code (or whatever it transpiles to).

## The outcome
This PR makes changes so only a single build process exists. The new process uses rollup to create 2 bundles. As a side effect a common bundle is created that contains the transpiled `store.ts` code that is common to the two bundles.
It also adds a `yarn validate` step to `yarn build` to make sure we're not building something with errors.

**Note**: ~~Because rollup does not bundle declaration files and we now out 2 bundles a file conflict occurs where the `.d.ts` files are generated twice, which fails the build. For this reason the exporting of declaration files was commented out and will be fixed in a later PR.~~ Declaration files are produced by running `tsc --emitDeclarationOnly` and then moving files to the correct place (namely `ConsentManagementPlatform.d.ts`). Unfortunately we can not use `tsconfig.json` for our settings in this case because we need that file for the rollup `tsc` build config, which is slightly different from our declaration config, so the parameters have to be supplied in the command line. This solution isn't the most elegant, but it does work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,73 +1,74 @@
 {
-  "name": "@guardian/consent-management-platform",
-  "version": "1.1.0-beta.4",
-  "description": "Library of useful utilities for managing consent state across *.theguardian.com",
-  "main": "lib/index.js",
-  "types": "lib/index.d.ts",
-  "scripts": {
-    "build": "yarn clean && tsc && rollup --config",
-    "clean": "rm -rf lib/*",
-    "test": "jest --config jestconfig.json",
-    "lint": "eslint src/**/*.ts src/**/*.tsx",
-    "tsc": "tsc --noEmit",
-    "validate": "yarn tsc && yarn lint && yarn test",
-    "fix": "yarn validate --fix",
-    "prepublishOnly": "yarn validate && yarn build"
-  },
-  "husky": {
-    "hooks": {
-      "pre-push": "yarn validate"
-    }
-  },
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,73 +1,74 @@
...
-    "rollup-plugin-commonjs": "^10.1.0",
-    "rollup-plugin-node-resolve": "^5.2.0",
-    "ts-jest": "^24.0.2",
-    "typescript": "^3.5.3"
-  },
-  "files": [
-    "lib/**/*"
-  ]
+    "name": "@guardian/consent-management-platform",
+    "version": "2.0.0",
```
ripecosta: Saving changed the tab spacing but here are the actual changes to this file:
- Version updated to `2.0.0`.
- Added `rollup-plugin-typescript2` dependency.
- Updated `typescript` dependency to `~3.5.3`

PR #65, opened 2020-01-27
Fix cookies not saving in viewer.gutools.co.uk

ripecosta:
Fixes a bug where cookies were not being saved when accessing viewer.gutools.co.uk.

This was happening because we never accounted for domains using a second level domain such as `viewer.gutools.co.uk`. So when slicing the full domain to remove subdomains we were getting `.co.uk` and setting it as the cookie domain, which the browser ignored.

### DIFF
--- a/src/cookies.test.ts
+++ b/src/cookies.test.ts
@@ -38,6 +38,11 @@ describe('Cookies', () => {
         domain: '.theguardian.com',
         expires: COOKIE_MAX_AGE,
     };
+    const previewCookieOptions = {
+        domain: '.gutools.co.uk',
+        expires: COOKIE_MAX_AGE,
+    };
+
     const iabConsentString = 'heL10W0rLd';
     const fakeNow = '12345';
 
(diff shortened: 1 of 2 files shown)


src/cookies.ts
```diff
@@ -2,6 +2,12 @@
 import Cookies from 'js-cookie';
 import { GuCookie, GuPurposeState } from './types';
 
+declare global {
+    interface Window {
+        guardian?: { config?: { page?: { isPreview?: boolean } } };
+    }
+}
+
```
GHaberis: Are we using this type declaration in anyway? Is `window.guardian.config.page.isPreview` something we can use to fix this bug? Or was that something you'd attempted?

ripecosta: No. That was from something else I was trying and forgot to remove.

src/cookies.test.ts
```diff
@@ -95,6 +105,33 @@ describe('Cookies', () => {
...
             );
         });
+
+        // TODO: update test when PECR purposes introduced
+        it('is in preview', () => {
+            Object.defineProperty(document, 'domain', {
+                value: 'viewer.gutools.co.uk',
+                configurable: true,
+            });
+            global.guardian = {
```
GHaberis: Do you need this `global.guardian` anymore?

ripecosta: I do not. Cleaned.

src/cookies.test.ts
```diff
@@ -95,6 +102,30 @@ describe('Cookies', () => {
                 cookieOptions,
             );
         });
+
+        // TODO: update test when PECR purposes introduced
+        it('is in preview', () => {
```
GHaberis: You think it's worth adding a test to check it saves the cookie against the correct domain with a subdomain ending in `.com`? eg: `profile.theguardian.com` should be against the `.theguardian.com`.

ripecosta: Discussed this point and reached the conclusion that the other tests already test for this case and they are sufficient.

PR #66, opened 2020-01-28
Fixes scrolling accessibility issue

ripecosta:
There is an accessibility issue on the modal for users without a scroll wheel. Users would have to click somewhere in the modal before being able to use the arrows keys to scroll. This PR forces the browser to focus on scrollable element of the modal immediately allowing the user to scroll the modal and also using tab to navigate it.

### DIFF
--- a/src/component/Modal.tsx
+++ b/src/component/Modal.tsx
@@ -203,6 +203,13 @@ class Modal extends Component<Props, State> {
         window.addEventListener('resize', () => {
             this.hideScrollbar();
         });
+
+        // This enables scrolling the modal using arrow keys as well
+        // as tabing through the items as soon as the modal is shown
+        const scrollableElem = document.getElementById(SCROLLABLE_ID);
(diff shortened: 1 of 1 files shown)


GHaberis:
To get around the issue of forcing the first collapsible button to be active (and therefore a different colour) could we try and `.focus()` a different element? The `div.#scrollable` can be made focussable if we add the attribute `tabindex="-1"`. `tabindex="-1"` allows elements besides the standard focussable elements to receive "programmatic" focus.

ripecosta:
> To get around the issue of forcing the first collapsible button to be active (and therefore a different colour) could we try and `.focus()` a different element? The `div.#scrollable` can be made focussable if we add the attribute `tabindex="-1"`. `tabindex="-1"` allows elements besides the standard focussable elements to receive "programmatic" focus.

I did not know that was possible. It would simplify the code as well. Will attempt and update the PR as appropriate. Thank you.

PR #67, opened 2020-01-31
Fix iOS 10 modal scrollable bug

GHaberis:
On iOS 10 devices (iPhone or iPad) the modal in the CMP UI is not entirely scrollable, users report being unable to scroll to the "Measurement" on/off radio buttons and the "Vendors" and "Features" lists without the scroll bouncing back to it's original position. I can verify this bug was reproducible on Browserstack.

The issue was caused by the "scrollable" div not rendering at the expected height, this PR fixes that by using a different technique to set the height of this element. 

This has been tested non Browserstack against a range of devices/browsers on desktop/mobile/tablet.

**Bug on iOS 10 iPhone...**

ios-10-buggy

**Fixed on iOS 10 iPhone...**

ios-10-fix

Trello card: https://trello.com/c/iDngBtE8/646-cmp-investigate-cmp-ios-1033-issues

### DIFF
--- a/src/component/CmpCollapsible.tsx
+++ b/src/component/CmpCollapsible.tsx
@@ -62,6 +62,7 @@ export class CmpCollapsible extends Component<Props, State> {
                                 css={css`
                                     flex-grow: 1;
                                     display: flex;
+                                    min-height: 25px;
                                 `}
                                 onClick={() => {
                                     this.toggleCollapsed();
--- a/src/component/Modal.tsx
+++ b/src/component/Modal.tsx
@@ -100,7 +100,6 @@ const copyContainerStyles = (headlineSerif: string) => css`
 
 const scrollableAreaStyles = (scrollbarWidth: number) => css`
     width: 100%;
-    height: 100%;
     overflow-y: scroll;
     -webkit-overflow-scrolling: touch;
     padding-right: ${scrollbarWidth}px; /* Increase/decrease this value for cross-browser compatibility */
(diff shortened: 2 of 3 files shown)


src/component/CmpCollapsible.tsx
```diff
@@ -62,6 +62,7 @@ export class CmpCollapsible extends Component<Props, State> {
                                 css={css`
                                     flex-grow: 1;
                                     display: flex;
+                                    min-height: 25px;
```
GHaberis: Completely unrelated but it fixes a misaligned "arrow" icon on the collapsible buttons in the CMP modal, that was driving me crazy :upside_down_face:

faresite: Love it when PR authors write these kind of comments, saves me time thinking about it ❤

src/component/Modal.tsx
```diff
@@ -111,8 +110,8 @@ const scrollableContainerStyles = (scrollbarWidth: number) => css`
     background-color: ${palette.neutral[100]};
     margin-right: -${scrollbarWidth}px;
     overflow: hidden;
-    max-height: 100%; /* For scrolling purposes */
-    height: 450px; /* For scrolling purposes */
+    display: flex;
+    flex-direction: column;
```
faresite: @GHaberis Just checking if this has been tested on other browsers like IE11?

GHaberis: Yep I've tested on IE 11 too 👍

PR #69, opened 2020-02-05
Update copy related to changing the privacy settings

ripecosta:
This PR updates the copy used to inform the user how they can change their consent settings for their current browser.


ripecosta:
> LGTM. Do copy changes to the CMP need sign-off outside the team?

Not this particular one, according to Alex.

PR #72, opened 2020-02-26
Show scrollbar in modal

GHaberis:
This PR reveals the scrollbar on the scrollable area of the `Modal` component. We had removed the scrollbar from the CMP UI back when it was a side banner to avoid having 2 scrollbars. When we moved to a bottom banner + modal UI we never revisited that decision but recent accessibility issues have caused us to rethink it. We have decided to reinstate it.

**Before:**
modal

**After:**
modal-scrollbar

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -25,6 +25,10 @@
         "@typescript-eslint/consistent-type-definitions": 0,
         "@typescript-eslint/explicit-function-return-type": 0,
         "@typescript-eslint/no-use-before-define": 0,
+        "class-methods-use-this": [
+            2,
+            { "exceptMethods": ["componentDidMount"] }
+        ],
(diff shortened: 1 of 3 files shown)


.eslintrc.json
```diff
@@ -25,6 +25,10 @@
         "@typescript-eslint/consistent-type-definitions": 0,
         "@typescript-eslint/explicit-function-return-type": 0,
         "@typescript-eslint/no-use-before-define": 0,
+        "class-methods-use-this": [
+            2,
+            { "exceptMethods": ["componentDidMount"] }
+        ],
```
GHaberis: Had to add this override as eslint was complaining that `componentDidMount` was a class method without a reference to `this` in it following the refactor.

PR #74, opened 2020-04-06
Move src-* dependencies into peerDependencies

GHaberis:
The Dotcom team identified that the CMP Component when bundled is bringing its own versions of `src-foundations`, `src-button` and `src-svgs` dependencies. These are already being bundled by DCR and other dependent packages

Since CMP is designed to be consumed by other applications,  these modules should be externalised, and specified as `peerDependencies`, rather than bundled with the CMP Component.

This PR does that by enacting the following changes:

- externalise src-* packages in the rollup config
- specify scr-* packages as peerDependencies rather than dependencies 
- Upgrade src-* packages to the latest version (0.16.1)
- `src-foundations@0.16.1` now exports `headlineSizes`/`bodySizes` as `px` values and not `rem` values, so this required an update to some styles.

After running a before/after bundle comparison on `dotcom-rendering` I found the change led to the desired drop in bundle size of ~ 3kb (compressed):

**Before (non-legacy dcr bundle)**
modern-old

**After (non-legacy drc bundle)**
modern-new

**Before (legacy dcr bundle)**
legacy-old

**After (legacy dcr bundle)**
legacy-new

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,15 +29,15 @@
     ],
     "license": "Apache-2.0",
     "dependencies": {
-        "@guardian/src-button": "^0.13.0",
-        "@guardian/src-foundations": "^0.13.0",
-        "@guardian/src-svgs": "^0.13.0",
         "consent-string": "^1.5.1",
         "js-cookie": "^2.2.1",
         "whatwg-fetch": "^3.0.0"
     },
     "peerDependencies": {
         "@emotion/core": "^10.0.21",
+        "@guardian/src-button": "^0.16.1",
+        "@guardian/src-foundations": "^0.16.1",
+        "@guardian/src-svgs": "0.16.1",
         "emotion-theming": "^10.0.27",
         "react": "^16.10.2"
     },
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@guardian/consent-management-platform",
-    "version": "2.0.10",
+    "version": "3.0.0",
```
GHaberis, line 3: Bumping this to new major version `3.0.0` as `peerDependencies` are part of the public API

src/component/Banner.tsx
```diff
@@ -106,12 +106,12 @@ const contentContainerStyles = (bodySerif: string) => css`
 `;
 
 const headlineStyles = (headlineSerif: string) => css`
-    font-size: ${headlineSizes.small}rem;
+    font-size: ${headlineSizes.small}px;
```
SiAdcock, line 109: You can use [`remHeadlineSizes`](https://github.com/guardian/source/blob/master/src/core/foundations/src/typography/index.ts#L94) if you prefer 🙂

GHaberis: Ah I missed that export! I'm going to stick with this for now as I think we only moved to `rem` as there wasn't a `px` export at the time, plus the final result is the same.

PR #75, opened 2020-04-09
Banner copy change

sndrs:
- [add link to vendors](https://trello.com/c/0h17K635)
  - doesn't open the list...
- [add text explaining that the CMP can be accessed at any time on first layer](https://trello.com/c/wD8xJF3K)
- [add text explaining access to information on their device is taking place by third parties in the 1st layer](https://trello.com/c/RuU8aGbA)

### DIFF
--- a/src/component/Banner.tsx
+++ b/src/component/Banner.tsx
@@ -106,12 +106,12 @@ const contentContainerStyles = (bodySerif: string) => css`
 `;
 
 const headlineStyles = (headlineSerif: string) => css`
-    font-size: ${headlineSizes.small}px;
+    font-size: ${headlineSizes.small}rem;
     font-weight: 700;
     line-height: 1.5;
 
     ${from.leftCol} {
-        font-size: ${headlineSizes.medium}px;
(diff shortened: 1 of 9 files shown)


sndrs:
Test is prepared in https://github.com/guardian/frontend/pull/22494

src/component/Banner.tsx
```diff
@@ -345,7 +417,7 @@ class Banner extends Component<Props, State> {
                                         }}
                                         aria-expanded={showInfo}
                                         aria-controls={INFO_LIST_ID}
-                                        tabIndex={1}
+                                        tabIndex={2}
```
GHaberis, line 420: I see the `tabIndex` has been increased by 1 because of the introduction of the `buttonAsLink` element which now occupies index 1 for users' in the `variant`, this makes sense. However if a user is in the `control` or not in the test at all this `buttonAsLink` element won't exist, so `tabIndex` 1 will be missing so I'd be I'd interested to know how the tabbing behaves.

sndrs: since they're in order, they behave the same

PR #76, opened 2020-04-17
Create dev task

sndrs:
- adds a dev server
- adds `yarn dev` to run the dev server
- commits yarn 1.22.4 to the repo


rollup.config.js
```diff
@@ -1,27 +1,34 @@
...
+        dir: 'lib',
+        format: 'cjs',
+    },
     external: [
         'react',
         '@emotion/core',
         '@guardian/src-button',
         '@guardian/src-foundations',
         '@guardian/src-svgs',
+        '@guardian/src-foundations/mq',
```
ripecosta, line 25: By setting these to be external don't we need to set them as peerDependencies?

sndrs: def, but `@guardian/src-foundations` is defined already and would include the modules that we use from within it
 https://github.com/guardian/consent-management-platform/blob/f0200337468091868eaed1d9885ff305c1f9056a/package.json#L36-L43

PR #78, opened 2020-04-23
IAB compliancy update

ripecosta:
This PR updates the copy on the banner and the behaviour when clicking the 'partners' link/button so address compliancy issues reported by the IAB.

When clicking the 'partners' link it will now open the modal with the Vendors section expanded and scroll to it, making it visible for the user.

Screenshot showing the updated copy:
Screenshot 2020-04-23 at 15 53 12

Also took the opportunity to remove some unused config variables.

### DIFF
--- a/src/component/Banner.tsx
+++ b/src/component/Banner.tsx
@@ -347,7 +347,7 @@ class Banner extends Component<Props, State> {
                                                 technologies – to improve your
                                                 experience on our site, analyse
                                                 how you use it, show you
-                                                personalised advertising. .
+                                                personalised advertising.
                                             </p>
                                             <p>
                                                 To find out more, read our{' '}
--- a/src/component/Banner.tsx
+++ b/src/component/Banner.tsx
@@ -287,7 +287,7 @@ interface Props {
     cookiePolicyUrl: string;
     iabPurposes: IabPurpose[];
     onEnableAllAndCloseClick: () => void;
-    onOptionsClick: () => void;
+    onOptionsClick: (shouldFocusVendors: boolean) => void;
     variant?: string;
 }
 
@@ -336,7 +336,7 @@ class Banner extends Component<Props, State> {
                                                 <button
                                                     css={buttonAsLinkStyles}
                                                     onClick={() => {
-                                                        onOptionsClick();
+                                                        onOptionsClick(true);
                                                     }}
                                                     tabIndex={1}
                                                 >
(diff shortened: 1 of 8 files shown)


src/component/CmpCollapsible.tsx
```diff
@@ -40,7 +41,7 @@ export class CmpCollapsible extends Component<Props, State> {
         super(props);
 
         this.state = {
-            collapsed: false,
+            collapsed: !!props.expandedByDefault, // TODO: collapsed has the wrong meaning. Should be the opposite.
```
ripecosta, line 44: I'll handle this issue in a future PR.

src/component/Modal.tsx
```diff
@@ -200,10 +202,17 @@ class Modal extends Component<Props, State> {
         if (scrollableElem) {
             (scrollableElem as HTMLElement).focus();
         }
+
+        if (this.props.focusVendors) {
+            const vendorsElem = window.document.getElementById(VENDORS_ID);
+            if (vendorsElem) {
+                vendorsElem.scrollIntoView({ behavior: 'smooth' });
```
ripecosta, line 209: Note: Setting the behaviour is currently only supported by Edge, Firefox, Chrome and Opera but polyfilling it means bloating our bundle. The behaviour on others browsers is that the window snaps to the Vendors. Seeing as only a small number of users surface the modal and this CMP will soon be replaced by Sourcepoint's I opted not to polyfill it.

jeteve: nice!

src/component/Banner.tsx
```diff
@@ -332,24 +332,22 @@ class Banner extends Component<Props, State> {
...
-                                                site, analyse how you use it,
-                                                show you personalised
-                                                advertising and share that
-                                                information with{' '}
+                                                We and our{' '}
                                                 <button
                                                     css={buttonAsLinkStyles}
                                                     onClick={() => {
-                                                        onOptionsClick();
+                                                        onOptionsClick(true);
```
jeteve, line 339: Maybe a comment to tell what true means?

src/component/CmpCollapsible.tsx
ripecosta: I want to keep the collapsed keyboard because the component is called CmpCollapsible and it makes it easier to understand for anyone looking at the code in the future.

src/component/Banner.tsx
ripecosta: The function signature is declared in line 290.

PR #94, opened 2020-05-20
Improve publising pipeline

sndrs:
- adds `yarn release` that uses [np](https://github.com/sindresorhus/np) for added security and convenience
- removes babel in favour of pure TS (babel wasn't necessary)
- creates ESM module version too 
- releases from `dist` rather than `lib` (more common name)
- updates docs


package.json
```diff
@@ -2,15 +2,15 @@
...
     "scripts": {
-        "build": "yarn validate && yarn build-only",
-        "build-only": "yarn clean && rollup --config && yarn declaration",
-        "clean": "rm -rf lib/*",
+        "build": "yarn clean && rollup --config && yarn build:types",
+        "clean": "rm -rf dist/*",
         "cy": "cypress open",
         "cy:run": "cypress run",
-        "declaration": "tsc --emitDeclarationOnly -d -m commonjs ",
+        "build:types": "tsc --declaration --emitDeclarationOnly --declarationDir 'dist'",
```
ripecosta, line 13: Non-blocking comment: The simplification of the library (not exposing a React component) in combination with the changes to the build process could mean we can go back to having the declaration flag in the tsconfig file and getting rid of this separate step to build types.

ripecosta, line 13: Non-blocking comment: The simplification of the library (not exposing a React component) in combination with the changes to the build process could mean we can go back to having the declaration flag in the tsconfig file and getting rid of this separate step to build types.

PR #99, opened 2020-06-01
Ccpa timing instrumentation

sndrs:
Add instrumenting to CMP. For existing tcfv1 CMP and sourcepoint CCPA, record:

- app init
- UI displayed
- user choice recovered

note if UI displayed the prob don't want to record user choice recovered


dev/app.tsx
```diff
@@ -3,9 +3,9 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import { ConsentManagementPlatform } from '../src/tcf/component/ConsentManagementPlatform';
-import { init } from '../src/index';
+import { init, onIabConsentNotification } from '../src/index';
 
-const initOptions = { useCcpa: true };
+const initOptions = { useCcpa: false };
```
ioanna0, line 8: Don't know much about the impact of this change, but assuming it was deliberate! Is it changing the default mode of the consent management in general?

ioanna0:
LGTM code wise 👍, just has a question on the logic change which I am not familiar with

dev/app.tsx
sndrs: it's was just left in there :)

it's the home page in the dev server that mimics a real world installation, so this flag mimics the value you'd get from checking the locale. it's only used in the dev server so shouldn't affect prod

PR #104, opened 2020-06-10
Add checkWillShowUI to API

ripecosta:
## What does this change?
Adds `checkWillShowUI` to API that will return the promise of a boolean. This boolean can be used to determine if a CCPA message will be shown on the current pageload.

### DIFF
--- a/dev/app.tsx
+++ b/dev/app.tsx
@@ -7,6 +7,7 @@ import {
     init,
     onIabConsentNotification,
     showPrivacyManager,
+    checkWillShowUI,
 } from '../src/index';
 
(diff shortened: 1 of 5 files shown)


ripecosta:
@sndrs 
> added a review by way of a PR – passing in the callback seems like a bit of indirection to me. what do you think about #106?

Thank you for your suggestions. Have taken parts of it and integrated into this branch. Please have another look when you have the chance.
I considered adding unit testing for this feature but would have had to complicate the solution a bit to accomplish that, which I felt wasn't worth the gain. Feels more like an end-to-end test, which we can implement later when we re-add cypress to the repo.

PR #134, opened 2020-06-22
Define CMP v4 API

ripecosta:
## What does this change?
This PR presents a suggested API for CMP v4 in README.md form. v4 of the CMP will replace our TCF v1 implementation with Sourcepoint's TCF v2 implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,9 +10,12 @@ The CMP applies the CCPA to users in the USA, and TCFv2 to everyone else.
 ## Usage
 
 ```js
-import { init, onConsentNotification } from '@guardian/consent-management-platform';
+import {
+    init,
+    onConsentNotification,
+} from '@guardian/consent-management-platform';
 
-init({ isInUS: true });
+init({ isInUsa: true });
 
 onConsentNotification(({ tcfState, ccpaState }) => {
     if (ccpaState) {
@@ -32,11 +35,11 @@ returns: `void`
 
 Adds the relevent privacy framework to the page. It needs to be run before any other API call.
 
-#### options.isInUS
+#### options.isInUsa
 
 type: `boolean`
 
-_Required!_ Declare whether your user is in the US or not.
+_Required!_ Declare whether your user is in the USA or not.
 
 ### onConsentNotification(callback)
 
@@ -57,7 +60,7 @@ If the consent state has already been acquired when `onConsentNotification` is c
 
 type: `Object` or `undefined`
 
-Reports the user's consent to the various TCFv2 uses:
+Reports the user's preferences to the TCFv2 purposes:
 
 ```js
 {
@@ -75,7 +78,7 @@ Reports whether user has withdrawn consent to sell their data:
 
 ```js
 {
-    doNotSell: Boolean
+    doNotSell: Boolean;
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -10,10 +10,7 @@ The CMP applies the CCPA to users in the USA, and TCFv2 to everyone else.
 ## Usage
 
 ```js
-import {
-    init,
-    onConsentNotification,
-} from '@guardian/consent-management-platform';
+import { init, onConsentNotification } from '@guardian/consent-management-platform';
 
 init({ isInUS: true });
 
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ onConsentNotification((tcfState, ccpaState) => {
 	if (tcfState !== null) {
 		console.log(tcfState); // { 1: true || false, 1: true || false, ... }
 	} else {
-		console.log(tcfState); // true || false
+		console.log(ccpaState); // true || false
 	}
 });
 ```
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ checkUiWillShow()
 
 returns: `void`
 
-Surfaces the relevant privacy manager, allowing the user to change their privacy preferences after the initial privacy message.
+Shows the relevant privacy manager, allowing the user to update their privacy preferences after they've closed the initial privacy message.
 
 **Example:**
 
--- a/README.md
+++ b/README.md
@@ -39,7 +39,9 @@ If `InitOptions.isInUS` is missing, `init` will default to running in TCF mode.
 
 `onConsentNotification(callback: ConsentCallack): void`
 
-When `onConsentNotification` is called it will add the supplied callback to a list of a callbacks. These will be fired when 1) after the consent state is first checked, and 2) the consent state changes (eg. if a user resurfaces the privacy manager and makes change to their privacy preferences). If the consent state is already known when `onConsentNotification` is called the callback is fired immediately. The signatures for the callback function and its parameters are:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,144 +1,112 @@
...
-##### source?: string
+```js
+import { onConsentNotification } from '@guardian/consent-management-platform';
+
+onConsentNotification((tcfState, ccpaState) => {
+	// Check whether it's in TCF or CCPA mode
+	if (tcfState !== null) {
+		console.log(tcfState); // { 1: true || false, 1: true || false, ... }
+	} else {
+		console.log(tcfState); // true || false
```
GHaberis, line 79: typo here I assume you mean log `ccpaState` in this else block

README.md
```diff
@@ -1,144 +1,104 @@
...
-onGuConsentNotification('functional', (functionalConsentState) => {
-	console.log(functionalConsentState); // true || false || null
-});
 ```
+interface InitOptions {
+	isInUS: boolean;
+}
+```
+
+If `InitOptions.isInUS` is missing, `init` will default to running in TCF mode.
```
sndrs, line 36: this feels dangerous to me – would it be safer to force people to be explicit?

ripecosta: I don't think it's that dangerous because in TCFv2 mode nothing fires unless given explicit consent but I'm happy to change it.
What do you suggest happens if the `isInUS` flag is missing?

sndrs: I was thinking that because we have to show the CCPA in California, we shouldn't give people the chance to not consider where their code will run?

how about we throw an error, as well as TS failing?

README.md
```diff
@@ -57,7 +60,7 @@ If the consent state has already been acquired when `onConsentNotification` is c
 
 type: `Object` or `undefined`
 
-Reports the user's consent to the various TCFv2 uses:
+Reports the user's preferences to the TCFv2 purposes:
```
sndrs, line 63: cheers :)

README.md
```diff
@@ -10,9 +10,12 @@ The CMP applies the CCPA to users in the USA, and TCFv2 to everyone else.
...
 
 ```js
-import { init, onConsentNotification } from '@guardian/consent-management-platform';
+import {
+    init,
+    onConsentNotification,
+} from '@guardian/consent-management-platform';
 
-init({ isInUS: true });
+init({ isInUsa: true });
```
sndrs, line 18: :+1:

README.md
```diff
@@ -75,7 +78,7 @@ Reports whether user has withdrawn consent to sell their data:
 
 ```js
 {
-    doNotSell: Boolean
+    doNotSell: Boolean;
```
sndrs, line 81: the example above uses js-style `,` – we should probably match them. want to go with TS version?

README.md
ripecosta: Throwing an error seems good since projects that don't use TS, such as frontend, won't see a TS error.

README.md
```diff
@@ -1,143 +1,120 @@
...
+### checkUiWillShow()
 
-The `onClose` property accepts a function, this will be executed once the user has submitted their consent, either via the clicking "I'm OK with that" button in the banner, or opening the options modal selecting their choices and clicking the "Save and close" button. You can add whatever logic you want in this function. Because the `ConsentManagementPlatform` component doesn't close itself a typical example of the logic that would be included in this function might be the updating of state to hide the `ConsentManagementPlatform` component.
+returns: `Promise<Boolean>`
 
-##### source?: string
+The returned promise resolves to `true` if the user will be shown the initial privacy message, or `false` if not.
 
-The `source` property accepts an optional string. The value passed to this will be sent to the consent logs once a user has submitted their consent. The value should indicate the site on which the CMP has been seen: eg. 'manage' for 'manage.theguardian.com'. The default value passed to the logs will be 'www'.
+If it's called before `init()`, it will `reject`.
```
sndrs, line 91: does this need to reject? what if it just didn't resolve until `init` has finished? if you never call `init` then `checkUiWillShow` never resolves?

README.md
ripecosta: The linter added this. But I supposed if it is a type definition TS makes more sense.

README.md
ripecosta: The idea behind rejecting is that it tells the developer using this package that they've forgot to call `init` beforehand. If `init` has been called, even if the initialisation process hasn't finished yet, it resolves. Not resolving it will not make it clear to the developer that something has gone wrong with their implementation.

README.md
```diff
@@ -39,38 +39,47 @@ Adds the relevent privacy framework to the page. It needs to be run before any o
 
 type: `boolean`
 
-_Required!_ Declare whether your user is in the USA or not.
+_Required! Throws error is not present._ Declare whether your user is in the USA or not.
```
sndrs, line 42:
```suggestion
_Required! Throws error if not present._ Declare whether your user is in the USA or not.
```

README.md
```diff
@@ -97,17 +106,17 @@ import { checkUiWillShow } from '@guardian/consent-management-platform';
...
+        console.log("checkUiWillShow() failed:", e); // "checkUiWillShow() failed: called before init()"
     );
 ```
 
 ### showPrivacyManager()
 
 returns: `void`
 
-Allows users to change their privacy preferences by (re-)showing the privacy manager after the initial privacy message.
+Surfaces the relevant privacy manager, allowing the user to change their privacy preferences after the initial privacy message.
```
sndrs, line 119: 'surfaces' reads a bit strangely here to me...

```suggestion
Shows the relevant privacy manager, allowing the user to update their privacy preferences after they've closed the initial privacy message.
```

README.md
```diff
@@ -1,143 +1,129 @@
...
-**Example:**
+Reports the user's relevant privacy preferences.
 
 ```js
-import { onIabConsentNotification } from '@guardian/consent-management-platform';
-
-onIabConsentNotification((iabConsentState) => {
-	console.log(iabConsentState); // { 0: true || false || null, 1: true || false || null, ... }
-});
+{
```
sndrs, line 62:
```suggestion
##### result

type: `Object` or `undefined`

{
```

README.md
```diff
@@ -1,143 +1,129 @@
...
+    tcfState?: Boolean,
+    ccpaState?: Boolean,
+}
 ```
 
-### CMP UI
+###### result.tcfState
```
sndrs, line 68:
```suggestion
##### result.tcfState
```

README.md
```diff
@@ -1,143 +1,129 @@
...
+    1: Boolean;
+    2: Boolean;
+    // etc
+}
 ```
 
-#### ConsentManagementPlatform React Component
-
-The properties the `ConsentManagementPlatform` component takes are listed below along with their Typescript definitions:
+###### result.ccpaState
```
sndrs, line 82:
```suggestion
##### result.ccpaState
```

README.md
```diff
@@ -1,143 +1,129 @@
...
+If the consent state has already been acquired when `onConsentNotification` is called, the callback will be invoked immediately.
 
-The keys in this object will match the IAB purpose IDs from the [IAB vendor list](https://vendorlist.consensu.org/vendorlist.json).
+#### callback(result)
 
-The `cmp` module will also listens for subsequent changes to the user's consent state (eg. if a user saves an update to their consent via the CMP modal), if this happens it will re-execute the callback, passing it a single argument, an object which reflects the latest consent granted to the IAB purposes.
+type: `function`
 
-**Example:**
+Reports the user's relevant privacy preferences.
```
sndrs, line 59:
```suggestion
Reports the user's relevant privacy preferences.

##### result

type: `Object`

```

PR #139, opened 2020-06-24
readme suggestions

sndrs:
readme suggestions. inline was getting really noisy so apologies for PR. 

i've tried to keep it as terse and generic as possible, see what you reckon

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,10 +10,7 @@ The CMP applies the CCPA to users in the USA, and TCFv2 to everyone else.
 ## Usage
 
 ```js
-import {
-    init,
-    onConsentNotification,
-} from '@guardian/consent-management-platform';
+import { init, onConsentNotification } from '@guardian/consent-management-platform';
 
 init({ isInUS: true });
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,104 +1,118 @@
 # Consent Management Platform
 
-Welcome to the Consent Management Platform, a library to manage privacy frameworks across \*.theguardian.com, from messaging to reporting the privacy state. All exports include Typescript definitions.
+> Consent management for `*.theguardian.com`.
 
-## What utilities does this offer?
+The CMP applies the CCPA to users in the USA, and TCFv2 to everyone else.
 
-### TCF and CCPA privacy frameworks
+![Types](https://img.shields.io/npm/types/@guardian/consent-management-platform)
```
ripecosta, line 7: This link takes me to an image. Is that right?

sndrs: github just does that for any images you add I think?

ripecosta: I thought you wanted to link it to a different page showing the type definition. Kind of like you link to our chat room below. Is that not the case?

sndrs: it was mainly just show the badge. is there a place it could link to?

ripecosta: Oh I see! All makes sense now. To my knowledge there isn't a place we can link to to show the type definitions. But I've been known to be wrong before 😜so I thought you knew something I didn't.

PR #141, opened 2020-06-24
strip debug from production build

sndrs:
## What does this change?

removes all `console` and `assert` calls from prod build

### DIFF
--- a/src/mark.ts
+++ b/src/mark.ts
@@ -1,8 +1,6 @@
-import { isGuardianDomain } from './domain';
-
 export const mark = (label: string) => {
 	window.performance?.mark?.(label);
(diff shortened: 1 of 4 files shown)


src/mark.ts
```diff
@@ -1,8 +1,6 @@
-import { isGuardianDomain } from './domain';
-
 export const mark = (label: string) => {
 	window.performance?.mark?.(label);
-	if (!isGuardianDomain() && process.env.NODE_ENV !== 'test') {
+	if (process.env.NODE_ENV !== 'test') {
```
ripecosta, line 3: Where are we setting `process.env.NOVE_ENV` to `test`? I can only see where we set it to `production` and to `development`

sndrs: jest does it automatically

PR #151, opened 2020-06-30
Stop CCPA stub being added to every page

buck06191:
## What does this change?
Only adds CCPA stub from Sourcepoint when CCPA is actually in use.
Reintroduces the code from #109 and puts the script element creation inside a function.

### CCPA
image

### TCF
image

### DIFF
--- a/src/ccpa/sourcepoint-stub.js
+++ b/src/ccpa/sourcepoint-stub.js
@@ -1,54 +1,55 @@
 /* eslint-disable */
-
-var e = false;
-var c = window;
-var t = document;
-function r() {
-    if (!c.frames['__uspapiLocator']) {
-        if (t.body) {
-            var a = t.body;
-            var e = t.createElement('iframe');
-            e.style.cssText = 'display:none';
-            e.name = '__uspapiLocator';
-            a.appendChild(e);
-        } else {
-            setTimeout(r, 5);
-        }
-    }
-}
-r();
-function p() {
-    var a = arguments;
-    __uspapi.a = __uspapi.a || [];
-    if (!a.length) {
-        return __uspapi.a;
-    } else if (a[0] === 'ping') {
-        a[2]({ gdprAppliesGlobally: e, cmpLoaded: false }, true);
-    } else {
-        __uspapi.a.push([].slice.apply(a));
-    }
-}
-function l(t) {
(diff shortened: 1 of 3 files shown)


src/ccpa/sourcepoint.ts
```diff
@@ -32,6 +32,13 @@ declare global {
 	}
 }
 
+const createCcpaStubScriptElement = (): HTMLElement => {
+	const ccpaStubSrc = `(function () { var e = false; var c = window; var t = document; function r() { if (!c.frames["__uspapiLocator"]) { if (t.body) { var a = t.body; var e = t.createElement("iframe"); e.style.cssText = "display:none"; e.name = "__uspapiLocator"; a.appendChild(e) } else { setTimeout(r, 5) } } } r(); function p() { var a = arguments; __uspapi.a = __uspapi.a || []; if (!a.length) { return __uspapi.a } else if (a[0] === "ping") { a[2]({ gdprAppliesGlobally: e, cmpLoaded: false }, true) } else { __uspapi.a.push([].slice.apply(a)) } } function l(t) { var r = typeof t.data === "string"; try { var a = r ? JSON.parse(t.data) : t.data; if (a.__cmpCall) { var n = a.__cmpCall; c.__uspapi(n.command, n.parameter, function (a, e) { var c = { __cmpReturn: { returnValue: a, success: e, callId: n.callId } }; t.source.postMessage(r ? JSON.stringify(c) : c, "*") }) } } catch (a) { } } if (typeof __uspapi !== "function") { c.__uspapi = p; __uspapi.msgHandler = l; c.addEventListener("message", l, false) } })();`;
```
jeteve, line 36: Hey :) You can actually run this code 'as is' , there should be no need to build a script tag with it :)

src/ccpa/sourcepoint-stub.js
```diff
@@ -1,54 +1,55 @@
...
-        }
-    } catch (a) {}
-}
-if (typeof __uspapi !== 'function') {
-    c.__uspapi = p;
-    __uspapi.msgHandler = l;
-    c.addEventListener('message', l, false);
-}
+export const runStub = () => {
+	var e = false;
```
jeteve, line 3: Maybe point to the source of the stub? It's not very clear from this code that it's a pretty printed version of the official one-liner.

PR #168, opened 2020-07-09
handle some messages not setting consent when dismissed

sndrs:
`onConsentChange` callbacks are also now invoked after the user dismisses banners that sourcepoint doesn't think of as setting consent (e.g. CCPA).

fixes https://trello.com/c/5XgaVhHG

### DIFF
--- a/src/types.d.ts
+++ b/src/types.d.ts
@@ -61,7 +61,7 @@ interface Window {
 			events?: {
 				onConsentReady: () => void;
 				onMessageReady: () => void;
-				onMessageReceiveData?: (data: { messageId: 0 | string }) => void;
+				onMessageReceiveData: (data: { messageId: 0 | string }) => void;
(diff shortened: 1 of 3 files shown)


src/types.d.ts
```diff
@@ -39,9 +59,13 @@ interface Window {
 				framework: 'tcfv2';
 			};
 			events?: {
-				onConsentReady?: () => void;
-				onMessageReady?: () => void;
+				onConsentReady: () => void;
+				onMessageReady: () => void;
 				onMessageReceiveData?: (data: { messageId: 0 | string }) => void;
```
mxdvl, line 64: Is this one optional? It isn't in `window._sp_ccpa` above.
```suggestion
				onMessageReceiveData: (data: { messageId: 0 | string }) => void;
```

sndrs: ooh no, nice one

PR #175, opened 2020-07-10
fix and update TCFv2 build

sndrs:
- installs old CMP with `@guardian` scope
- stops trying to build the old CMP
- adds an ES module version to the package
- removes the IAB tcf stub for the sourcepoint one (hopefully less likely to get bitten)
- cleans up `package.json`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,49 +1,34 @@
 {
 	"name": "@guardian/consent-management-platform",
 	"version": "3.4.8",
-	"description": "Library of useful utilities for managing consent state across *.theguardian.com",
+	"description": "Consent management for *.theguardian.com.",
+	"homepage": "https://github.com/guardian/consent-management-platform.git",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -1,50 +1,34 @@
...
 	"main": "dist/index.js",
+	"module": "dist/index.esm.js",
 	"types": "dist/index.d.ts",
 	"files": [
 		"dist/**/*"
 	],
 	"scripts": {
-		"build": "yarn validate && yarn build-only",
-		"build-only": "yarn clean && rollup --config && yarn declaration",
+		"build": "npm-run-all clean --parallel compile types",
```
mxdvl, line 15: Mixing NPM & Yarn?

sndrs: yeah it's a confusing name. it allows us to run scripts in parallel and serial as we choose. in this instance, `clean` runs, then `compile` and `types` are run in parallel, which i don't think yarn can do?

PR #181, opened 2020-07-14
TCFv2 expose old CCPA init()

mxdvl:
## What does this change?

While the AB test is running, we will still need to expose the old CMP's `init()` for users in the USA.
The three 3, mutually-exclusive groups—thanks @ripecosta for clarifying—that a user might fall in are:

1. In USA: old CCPA _(requires `init({ccpaApplies: true})`)_
2. In RotW: old TCFv1 _(no `init()` required)_
3. In RotW (1%): new TCFv2 _(new `init({isInUsa: false})`)_

The old CMP's init is required for **Group 1**, and this PR aims to fix that. 

The TCFv2 AB test will run smoothly. 

## Further notes

In frontend, the CMP will then be called through [`guardian/frontend/…/javascripts/boot.js#36`](https://github.com/guardian/frontend/blob/193bc01c52fe773d449d86604feba027e29ab6e2/static/src/javascripts/boot.js#L36)  :

```js
// Start CMP
if (isCcpaApplicable()) {
  oldCmp.init({ useCcpa: true });
} else if (isInTcfv2Test()) {
  cmp.init({ isInUsa: isInUsa() });
}
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,7 @@
 /* eslint-disable no-console */
 
 import {
+	init,
 	onGuConsentNotification,
 	onIabConsentNotification,
 	shouldShow,
@@ -48,6 +49,7 @@ export const cmp = {
 export { onConsentChange } from './onConsentChange';
 
 export const oldCmp = {
+	init,
 	onGuConsentNotification,
 	onIabConsentNotification,
 	shouldShow,
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,7 +1,7 @@
 /* eslint-disable no-console */
 
 import {
-	init,
+	init as oldInit,
 	onGuConsentNotification,
 	onIabConsentNotification,
 	shouldShow,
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -1,11 +1,6 @@
 /* eslint-disable no-console */
 
-import {
-	onGuConsentNotification,
-	onIabConsentNotification,
-	shouldShow,
-} from '@guardian/old-cmp';
-import { ConsentManagementPlatform } from '@guardian/old-cmp/dist/ConsentManagementPlatform';
+import { oldCmp as olderCmp } from './oldCmp';
```
mxdvl, line 3: Is there a way to directly import the old CMP into an object, without having two `init` in the same scope?

mxdvl: Yes, [you just need the curly braces around the exported object](https://javascript.info/import-export):

```js
import { oldCmp } from './oldCmp';
export { oldCmp };
```

mxdvl:
Merged into -> https://github.com/guardian/consent-management-platform/pull/156

PR #188, opened 2020-07-23
TCFv2 fix Sourcepoint IDs

mxdvl:
## What does this change?

Update Sourcepoint ID so they pointing to the right privacy managers and vendor lists.

## How to test

On PROD / CODE, a place in the US should be able to see the privacy manager when clicking "Do Not Sell My Data"

We should be able to see the privacy manager in CCPA (US).

Yes: Sourcepoint needs to work!

## Images

n/a

### DIFF
--- a/src/lib/sourcepointConfig.ts
+++ b/src/lib/sourcepointConfig.ts
@@ -1,0 +1,3 @@
+export const ACCOUNT_ID = 1257;
+export const VENDOR_LIST_ID_CCPA = '5eba7ef78c167c47ca8b433d';
+export const CMP_ID_TCFV2 = 106842;
--- a/src/lib/sourcepointConfig.ts
+++ b/src/lib/sourcepointConfig.ts
@@ -1,3 +1,3 @@
 export const ACCOUNT_ID = 1257;
-export const VENDOR_LIST_ID_CCPA = '5eba7ef78c167c47ca8b433d';
-export const CMP_ID_TCFV2 = 106842;
+export const PRIVACY_MANAGER_CCPA = '5eba7ef78c167c47ca8b433d';
+export const PRIVACY_MANAGER_TCFV2 = 106842;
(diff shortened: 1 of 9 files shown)


mxdvl:
Hi @ripecosta, do you think it would be worth to have all the Sourcepoint IDs in their own file and import that, or is the current method of extracting it at the top of the file enough? And should there be an option to override these at `init()`?

ripecosta:
> Hi @ripecosta, do you think it would be worth to have all the Sourcepoint IDs in their own file, like the current `lib/accountId .js`, and import them this way, or is the current method of extracting it as a constant at the top of the file better?
> 
> It's worth noting that an update to these Sourcepoint IDs would require a new package release in the future. I think that's better than passing them as arguments to `init()`.

Ordinarily I would say no because these are only being used in a single place. However because you already have the pattern of a separate file for the `ACCOUNT_ID`(that one is used in multiple places), I believe it's better to keep all the IDs in the same file and rename it to something like `sourcepoint_config.js`. That way if you need to change any of them in the future they'll all be in the same place instead of scattered around different files in the code.

Agree with your point that if these IDs change it's better to release a new version rather than passing these to `init()`.

src/lib/sourcepointConfig.ts
```diff
@@ -0,0 +1,3 @@
+export const ACCOUNT_ID = 1257;
+export const VENDOR_LIST_ID_CCPA = '5eba7ef78c167c47ca8b433d';
+export const CMP_ID_TCFV2 = 106842;
```
ripecosta, line 3: These variable names could be more descriptive.
```suggestion
export const CCPA_PRIVACY_MANAGER_ID = '5eba7ef78c167c47ca8b433d';
export const TCFV2_VENDORLIST_ID = 106842;
```

mxdvl: Agreed, updated now.

PR #190, opened 2020-07-28
TCFv2 return all states

mxdvl:
## What does this change?

This PR returns a consistent `TCData.purpose.consents` object, where all purposes (1 to 10) are set to a boolean.

The native IAB implementation would return an empty object `{}` in case all purposes are false. This is due to their implementation using a [Boolean Vector]. In practice, this would mean that previously:

- consented purposes: `[]` -> consent object: `{}`
- consented purposes: `[2]` -> consent object: `{ '1': false, '2': true }`
- consented purposes: `[2,5]` -> consent object: `{'1': false, '2':true, '3': false, '4': false, '5': true }`

The new behaviour is now more explicit:

- consented purposes: `[]` -> consent object: `{'1': false, '2': false, '3': false, … , '9': false, '10': false}`
- consented purposes: `[2]` -> consent object: `{ '1': false, '2': true, '3': false, … , '9': false, '10': false }`
- consented purposes: `[2,9]` -> consent object: `{'1': false, '2': true, '3': false, … , '9': true, '10': false }`

[Boolean Vector]: https://github.com/InteractiveAdvertisingBureau/iabtcf-es/blob/fd137351359c346ce2eb45ecc4408909bebf8c41/modules/cmpapi/src/response/TCData.ts#L171

Less errors when checking consent. For example through `consents.every(Boolean)`.

This should help reduce bugs.

## Images

n/a

### DIFF
--- a/src/onConsentChange.test.js
+++ b/src/onConsentChange.test.js
@@ -1,6 +1,6 @@
 /* eslint-disable no-underscore-dangle */
 import waitForExpect from 'wait-for-expect';
-import { onConsentChange, invokeCallbacks } from './onConsentChange';
+import { onConsentChange, invokeCallbacks, _ } from './onConsentChange';
 
 const uspData = {
 	version: 1,
@@ -62,20 +62,12 @@ describe('onConsentChange', () => {
 	});
 
 	it('returns all 10 TCF purposes even if they are false', () => {
-		const consents = {
-			'1': false,
-			'2': false,
-			'3': false,
-			'4': false,
-			'5': false,
-			'6': false,
-			'7': false,
-			'8': false,
-			'9': false,
-			'10': false,
-			...tcfData.purposes.consents,
-		};
+		const consents = _.fillAllConsents(tcfData.purposes.consents);
 
 		expect(Object.keys(consents)).toHaveLength(10);
+		expect(consents[1]).toEqual(true);
+		expect(consents[3]).toEqual(true);
+		expect(consents[9]).toEqual(false);
+		expect(consents[10]).toBeDefined();
 	});
 });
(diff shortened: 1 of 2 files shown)


src/onConsentChange.ts
```diff
@@ -79,7 +79,19 @@ const getConsentState: () => Promise<ComparedConsentState> = () => {
 
 			Promise.all([getTCDataPromise, getCustomVendorConsentsPromise])
 				.then((data) => {
-					const { consents } = (data[0] as TCFData).purpose;
+					const consents = {
```
ioanna0, line 82: Why we are hardcoding 1-10 to false? Is it like their initial state or something? Is there a way to use the tcf response to populate their value?

ioanna0: Ah think i got it, is it because purpose.consents keys will always be the same and thus override their default value, false?

src/onConsentChange.test.js
```diff
@@ -51,4 +60,22 @@ describe('onConsentChange', () => {
...
+			'6': false,
+			'7': false,
+			'8': false,
+			'9': false,
+			'10': false,
+			...tcfData.purposes.consents,
+		};
+
+		expect(Object.keys(consents)).toHaveLength(10);
+	});
```
ioanna0, line 80: Maybe few more expectations will help
```suggestion
	expect(consents['1']).toBe(true);
	expect(consents['2']).toBe(false);
	expect(consents['3']).toBe(true);
	});
```

mxdvl: Technically, this should also be using a "real" test. I did not manage to mock the `__tcfapi` properly, however.

mxdvl: I've moved this into its own function call and tested that instead!

ioanna0: Nice one !  👍

PR #197, opened 2020-08-07
Add pubData to SourcePoint config object

janua:
## What does this change?

This adds a pubData type to the `window._sp_.config` type and also updates the `init` function so that we can pass along data to be used by SourcePoint. This will allow us on web to associate the current `browserId` with an impression/consent signal from the CMP.

The `pubData` is a free field where we can add anything; it will show up in the data that SourcePoint then send to us.

@guardian/commercial-dev

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,12 @@ type: `boolean`
 Declare whether your user is in the USA or not. Required – *throws an error if
 it's missing.*
 
+#### options.pubData
+
+type: `Object`
+
+Pass additional parameters for for reporting. Optional.
+
 #### Example
 
 ```js
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Pass additional parameters for for reporting. Optional.
 #### Example
 
 ```js
-cmp.init({ isInUsa: false });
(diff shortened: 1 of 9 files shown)


mxdvl:
@janua is there a reason why we don't pass this to the CCPA sourcepoint window object?

See `window._sp_ccpa`  in https://github.com/guardian/consent-management-platform/blob/TCFv2/src/ccpa/sourcepoint.ts

README.md
```diff
@@ -30,6 +30,12 @@ type: `boolean`
 Declare whether your user is in the USA or not. Required – *throws an error if
 it's missing.*
 
+#### options.pubData
```
ripecosta, line 33: Would be nice to show how to use this object by updating the example given in line 42 to include how to pass the browserId. Doesn't need to be part of this PR though.

src/types.d.ts
```diff
@@ -58,6 +63,9 @@ interface Window {
 			targetingParams: {
 				framework: 'tcfv2';
 			};
+			pubData: {
+				browserId?: string;
```
mxdvl, line 67: You may want to use the same `pubData` interface you're using on `_sp_ccpa`?

PR #202, opened 2020-08-11
Reuse PubData type

janua:
## What does this change?

Minor change to reuse the `PubData` type within the TCF `_sp_` object.

@guardian/commercial-dev @mxdvl


janua:
I now have moved the `PubData` type to the top of the file; DCR isn't happy with it's location when importing it

PR #204, opened 2020-08-12
Tcfv2 types (take2)

sndrs:
export types (hopefully!) 🤞


mxdvl:
For posterity, we should add some details on the decisions behind this.

sndrs:
we didn't fully understand the relationship between definition files and source files: 

- definition files (`*.d.ts`) describe the source code (`*.ts`) they accompany
- they are nothing to do with anything that sourcecode might generate. 

to make types available to consumers of your built application, you need to put them in source code, since the compiler will create whatever definition files your final bundle will need from that, not your local definition files.

PR #206, opened 2020-08-13
commercial.dcr.js hotfix

sndrs:
## What does this change?
- puts all exports on window and uses them if they exist

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -25,8 +25,8 @@ function init({ pubData, isInUsa }: { pubData?: PubData; isInUsa: boolean }) {
 	if (window) {
 		window.guardian = {
 			cmp: {
-				initialised: true,
 				...window.guardian?.cmp,
+				initialised: true,
 			},
 		};
 	}
--- a/src/index.ts
+++ b/src/index.ts
@@ -18,17 +18,13 @@ const initialised = new Promise((resolve) => {
 
 function init({ pubData, isInUsa }: { pubData?: PubData; isInUsa: boolean }) {
 	// *************** START commercial.dcr.js hotfix ***************
-	if (window?.guardian?.cmp?.initialised) {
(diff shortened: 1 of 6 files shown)


src/index.ts
```diff
@@ -13,6 +17,18 @@ const initialised = new Promise((resolve) => {
...
 
 function init({ pubData, isInUsa }: { pubData?: PubData; isInUsa: boolean }) {
+	// *************** START commercial.dcr.js hotfix ***************
+	if (window?.guCmpHotFix?.initialised) {
+		return;
+	}
+
+	if (window) {
+		window.guCmpHotFix = {
+			initialised: true,
```
sndrs, line 27:
```suggestion
			...window.guCmpHotFix,
			initialised: true,
```

src/index.ts
```diff
@@ -35,12 +51,29 @@ function showPrivacyManager() {
...
+
+// export { onConsentChange } from './onConsentChange';
+
+const actualExports = {
+	cmp: { init, willShowPrivacyMessage, showPrivacyManager },
+	onConsentChange: actualOnConsentChange,
 };
 
-export { onConsentChange } from './onConsentChange';
+if (window) {
```
jeteve, line 68: Shouldn't that be part of the init function to avoid assigning `guCmpHotFix` more than once?

mxdvl: There's a `window.guCmpHotFix.initialised` value set on L28 to check it only gets initialised once.

jeteve: I'm not sure that protects against double initialisation in the case the code is loaded as part of two different JS 
 bundles, as this window assignment is globally scoped?

mxdvl: You're right, and that's why we spread using the spread operator below: `...window.guCmpHotFix`

PR #211, opened 2020-08-18
Fix (hopefully) ads in USA

sndrs:
After good sleuthing by @jeteve, we noticed that if `onConsentChange` is called _while_ `cmp.init` is being run, callbacks that have already been added may never be invoked 😱…

## What does this change?

- use a per-callback state comparison, rather than a global one
- moves the first `invokeCallbacks` for CCPA to immediately after the `uspapi` is available

### DIFF
--- a/src/ccpa/sourcepoint.ts
+++ b/src/ccpa/sourcepoint.ts
@@ -14,6 +14,9 @@ export const willShowPrivacyMessage = new Promise<boolean>((resolve) => {
 export const init = (pubData = {}) => {
 	stub();
 
+	// invoke callbacks ASAP in USA
+	invokeCallbacks();
+
 	// make sure nothing else on the page has accidentally
 	// used the _sp_* name as well
 	if (window._sp_ccpa) {
@@ -47,16 +50,6 @@ export const init = (pubData = {}) => {
 				},
 				onMessageReceiveData: (data) => {
 					resolveWillShowPrivacyMessage?.(data.msg_id !== 0);
-				},
-				onMessageChoiceSelect: (_choiceId, choiceTypeID) => {
-					if (
-						// https://documentation.sourcepoint.com/web-implementation/sourcepoint-set-up-and-configuration-v2/optional-callbacks#choice-type-id-descriptions
-						choiceTypeID === 11 ||
-						choiceTypeID === 13 ||
-						choiceTypeID === 15
-					) {
-						setTimeout(invokeCallbacks, 0);
-					}
 				},
 			},
 		},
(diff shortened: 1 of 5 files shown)


src/ccpa/sourcepoint.ts
```diff
@@ -14,6 +14,9 @@ export const willShowPrivacyMessage = new Promise<boolean>((resolve) => {
 export const init = (pubData = {}) => {
 	stub();
 
+	// invoke callbacks ASAP in USA
+	invokeCallbacks();
```
mxdvl, line 18: This is okay, because it's an _opt-out_ framework.

src/onConsentChange.ts
```diff
@@ -55,27 +60,31 @@ const getConsentState: () => Promise<ComparedConsentState> = () => {
...
 					reject();
 				}
 			});
 		} else if (window.__tcfapi) {
 			// in RoW - https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20CMP%20API%20v2.md
-			const tcfApi = window.__tcfapi;
 
 			const getTCDataPromise = new Promise((subResolve, subReject) => {
-				tcfApi('getTCData', 2, (tcfData, success) => {
+				window.__tcfapi?.('getTCData', 2, (tcfData, success) => {
```
mxdvl, line 72: Cleaner!

src/onConsentChange.ts
```diff
@@ -131,31 +138,17 @@ const fillAllConsents: ConsentObject = (consentVector) => {
...
 	// if consentState is already available, invoke callback immediately
 	getConsentState()
 		.then((consentState) => {
-			if (consentState) callBack(consentState.state);
+			if (consentState) {
+				invokeCallback(newCallback, consentState);
+			}
 		})
 		.catch(() => {
 			// do nothing - callback will be added the list anyway
```
mxdvl, line 154:
```suggestion
			// do nothing - callback will be added the list anyway and executed when consent changes
```

src/ccpa/sourcepoint.ts
```diff
@@ -48,16 +51,6 @@ export const init = (pubData = {}) => {
 				onMessageReceiveData: (data) => {
 					resolveWillShowPrivacyMessage?.(data.msg_id !== 0);
 				},
-				onMessageChoiceSelect: (_choiceId, choiceTypeID) => {
```
mxdvl, line 51: Why is this being removed entirely?

sndrs: it's fired when a user clicks on the privacy message (dismissed it without making a choice etc). since we're running callbacks as soon as the `uspapi` is available we don't need to handle that case now. i suspect this may be true for TCF too, but we can look at that after

PR #223, opened 2020-08-21
1st pass at minifying the output

sndrs:
### Before
```
index.js: 18.64 kB → 5.11 kB (gzip)
index.esm.js: 18.55 kB → 5.07 kB (gzip)
```

### After
```
index.js: 7.67 kB → 2.75 kB (gzip)
index.esm.js: 7.61 kB → 2.73 kB (gzip)
```

to do:

- [ ] check terser settings


sndrs:
there's no point to this :)

PR #226, opened 2020-08-25
Restore old CMP

sndrs:
## What does this change?

adds `oldCmp` back so we can proceed with improvements to CMP while working out how to remove old cmp from dotcom

cc @ripecosta – back from the grave!

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,176 @@ onConsentChange(({ tcfv2, ccpa }) => {
 });
 ```
 
+## TCFv1 (deprecated)
+
+_These should not be used after 15 August 2020, and will be deleted shortly afterwards._
+
+#### oldCmp.onGuConsentNotification
+
+This function takes 2 arguments, the first is the purpose name (a `string`)
+that is relevant to the code you're running eg. "functional" OR "performance",
+and the second is a callback (a `function`).
+
+When `oldCmp.onGuConsentNotification` is called it will execute the callback
+immediately, passing it a single argument (a `boolean` or `null`) which
+indicates the user's consent state at that time for the given purpose name.
+
+The `cmp` module also listens for subsequent changes to the user's consent state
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -27,6 +27,9 @@
 		"types": "tsc --project ./tsconfig.build.json",
 		"validate": "npm-run-all --parallel tsc lint 'test --onlyChanged'"
 	},
+	"dependencies": {
+		"@guardian/old-cmp": "npm:@guardian/consent-management-platform@^3.4.10"
```
mxdvl, line 31: Be aware 3.4.10 is compatible with Source v2.0!

src/index.ts
```diff
@@ -79,3 +79,5 @@ if (window) {
 export const { cmp, onConsentChange } =
 	(window?.guCmpHotFix as typeof actualExports) || actualExports;
 // *************** END commercial.dcr.js hotfix ***************
+
+export { oldCmp } from './oldCmp';
```
mxdvl, line 83:
```suggestion
export { oldCmp } from './oldCmp'; // DEPRECATED: will be tree-shaken out
```

README.md
```diff
@@ -155,6 +155,176 @@ onConsentChange(({ tcfv2, ccpa }) => {
 });
 ```
 
+## TCFv1 (deprecated)
+
+_These should not be used after 15 August 2020, and will be deleted shortly afterwards._
```
mxdvl, line 160:
```suggestion
_Since 15 August 2020, TCFv1 is deprecated and `oldCmp` should not be used._
```

mxdvl (approved):

package.json
sndrs: this is the version currently shipping in `4.0.*` – we should probably match that?

mxdvl: Great!

PR #233, opened 2020-08-27
Fix coverage reporting take 2

sndrs:
adds test coverage reporting to PRs.

this will need some configuring to make it as useful as possible, but we need a baseline in master to do that.


mxdvl (approved):
👍 let's add a baseline in `main`

PR #248, opened 2020-09-01
Tidying up take 2

sndrs:
## What does this change?

- move code around to make it more testable and (hopefullly) clearer to use 

_re-animation of #217_

PR #255, opened 2020-09-02
Fix types build

sndrs:
## What does this change?

- fixes regression in #248 
- adds `build` CI step to make sure branches don't break the build (like #248 did)


mxdvl (commented):
👍

PR #263, opened 2020-09-07
Package version check

mxdvl:
## What does this change?

This adds a version check on the `window.guCmpHotFix`, in order to confirm whether two different versions are running at the same time. This is a follow-up to #206 and #261.

- Move replace plugin before babel

## How to test

If two different versions of the CMP are loaded, for example one from frontend and one from DCR, there should be a warning in the console.

Errors due to incompatible CMP package versions will be much easier to catch.

This does not reveal any privileged information about The Guardian’s systems.

## Images

n/a


src/index.ts
```diff
@@ -62,6 +63,7 @@ export const cmp = (window.guCmpHotFix.cmp ||= {
 	init,
 	willShowPrivacyMessage,
 	showPrivacyManager,
+	version: __PACKAGE_VERSION__,
```
sndrs, line 66: i think you'll need to declare this as a global

src/types/window.d.ts
```diff
@@ -17,6 +17,7 @@ declare global {
 				}) => void;
 				willShowPrivacyMessage: WillShowPrivacyMessage;
 				showPrivacyManager: () => void;
+				version: string;
```
sndrs, line 20: i wonder if this should be at the top level i.e. `window.guCmpHotFix.version`?

mxdvl: Sure, that makes sense.

src/index.ts
mxdvl: TIL about globals in Jest, ESLint & TS: e511a63

src/index.test.js
```diff
@@ -12,16 +13,24 @@ jest.mock('./ccpa', () => ({
...
+	willShowPrivacyMessage: jest.spyOn(actualTCFv2, 'willShowPrivacyMessage'),
+};
+
+// jest.mock('./tcfv2', () => ({
+// 	TCFv2: {
+// 		init: jest.fn(),
+// 		showPrivacyManager: jest.fn(),
+// 		willShowPrivacyMessage: () => Promise.resolve('iwillshowit'),
+// 	},
+// }));
```
mxdvl, line 28:
```suggestion
```

src/index.test.js
```diff
@@ -12,16 +13,16 @@ jest.mock('./ccpa', () => ({
...
-		init: jest.fn(),
-		showPrivacyManager: jest.fn(),
-		willShowPrivacyMessage: () => Promise.resolve('iwillshowit'),
-	},
-}));
+const TCFv2 = {
+	init: jest.spyOn(actualTCFv2, 'init'),
+	showPrivacyManager: jest.spyOn(actualTCFv2, 'showPrivacyManager'),
+	willShowPrivacyMessage: jest.spyOn(actualTCFv2, 'willShowPrivacyMessage'),
+};
```
mxdvl, line 20: Here’s we can use the `spyOn` method, @sndrs.

src/index.test.js
```diff
@@ -67,6 +68,21 @@ describe('hotfix cmp.init', () => {
 		expect(window.guCmpHotFix.initialised).toBe(true);
 	});
 
+	it('warn if two versions are running simultaneously', () => {
+		const spy = jest.spyOn(global.console, 'warn');
+		cmp.init({ isInUsa: false });
+		const cmpVersion = [window.guCmpHotFix.cmp.version, 'mockedVersion'];
+		// eslint-disable-next-line prefer-destructuring
+		window.guCmpHotFix.cmp.version = cmpVersion[1];
```
mxdvl, line 76:
```suggestion
		const cmpVersions = [window.guCmpHotFix.cmp.version, 'mockedVersion'];
		// eslint-disable-next-line prefer-destructuring
		window.guCmpHotFix.cmp.version = cmpVersions[1];
```

src/index.test.js
```diff
@@ -67,6 +68,21 @@ describe('hotfix cmp.init', () => {
...
+
+		cmp.init({ isInUsa: false });
+
+		expect(spy).toHaveBeenCalledWith(
+			'Two different versions of the CMP are running:',
+			cmpVersion,
```
mxdvl, line 82:
```suggestion
			cmpVersions,
```

src/index.test.js
sndrs: looks great! should we do the same for CCPA?

src/index.test.js
```diff
@@ -12,16 +13,16 @@ jest.mock('./ccpa', () => ({
...
 }));
 
-jest.mock('./tcfv2', () => ({
-	TCFv2: {
-		init: jest.fn(),
-		showPrivacyManager: jest.fn(),
-		willShowPrivacyMessage: () => Promise.resolve('iwillshowit'),
-	},
-}));
+const TCFv2 = {
```
sndrs, line 16: maybe this should be called `TCFv2Spy` and then the import can be the default name?

src/index.test.js
```diff
@@ -67,6 +68,21 @@ describe('hotfix cmp.init', () => {
 		expect(window.guCmpHotFix.initialised).toBe(true);
 	});
 
+	it('warn if two versions are running simultaneously', () => {
+		const spy = jest.spyOn(global.console, 'warn');
+		cmp.init({ isInUsa: false });
+		const cmpVersions = [window.guCmpHotFix.cmp.version, 'mockedVersion'];
+		// eslint-disable-next-line prefer-destructuring
+		window.guCmpHotFix.cmp.version = cmpVersions[1];
```
sndrs, line 76: I wasn't quite clear what was going on here. does this capture it?

```suggestion

		const currentVersion = window.guCmpHotFix.cmp.version;
		window.guCmpHotFix.cmp.version = 'mockVersion';
```

src/index.test.js
```diff
@@ -67,6 +68,21 @@ describe('hotfix cmp.init', () => {
...
+
+		cmp.init({ isInUsa: false });
+
+		expect(spy).toHaveBeenCalledWith(
+			'Two different versions of the CMP are running:',
+			cmpVersions,
+		);
```
sndrs, line 83: _only needed if you go with the above suggestion..._
```suggestion
		expect(
			spy,
		).toHaveBeenCalledWith('Two different versions of the CMP are running:', [
			currentVersion,
			window.guCmpHotFix.cmp.version,
		]);
```

sndrs (changes requested):

src/index.test.js
mxdvl: I don’t mind either way, but it made more sense to do it this way to make it work the same way as mocks, meaning you’re calling a function with the same name as you would normally.

Calling `TCFv2.init` behaves as you would expect, as it is calling `actualTCFv2.init`.

src/index.test.js
```diff
@@ -67,6 +68,21 @@ describe('hotfix cmp.init', () => {
 		expect(window.guCmpHotFix.initialised).toBe(true);
 	});
 
+	it('warn if two versions are running simultaneously', () => {
+		const spy = jest.spyOn(global.console, 'warn');
```
sndrs, line 72: if this was mocked, i think it would stop it showing in the jest output while still making the same check?

src/index.test.js
mxdvl: I was wondering which way was clearer. More explicit is better!

src/index.test.js
mxdvl: Yes, good idea!

src/index.test.js
mxdvl: I’m not sure `willShowPrivacyMessage` is ever called, then. I think this is one for the e2e test?

sndrs (approved):
nice one 👍

sndrs:
actually, just occurred to me – does this strip out the `console.warn`?

https://github.com/guardian/consent-management-platform/blob/a08a21a042dc59f04539a8fcc6a0333c17734296/rollup.config.js#L34-L37

mxdvl:
@sndrs this is an excellent point. There was another warning stripped, too:

> cmp.showPrivacyManager() was called before the CMP was initialised. This will work but you are probably calling cmp.init() too late.

For the time being, I am ignoring the index file with `exclude: ['src/index.*']`: 835e94b

PR #267, opened 2020-09-08
Add Cypress tests

mxdvl:
## What does this change?

Provide end-to-end (e2e) testing with Cypress. This will help us catch changes and errors on live pages, as well as compare screenshots of the visual appearance of the CMP.

**TODO:**

- [x] Add interactions with the CMP
- [ ] Enable testing in USA
- [x] Update CI checks?

Previously attempted in #87.

## How to test

The beauty of it is you don’t have to. You can run it locally with `yarn test:integration:visual`. 

Improved DevX and more bugs caught earlier.

Changes in the Sourcepoint library may create false negatives.

## Images

image


mxdvl:
Cypress probably fails in CI because the servers are in the US:

This is the result of `tcData` in `/src/tcfv2/getConsentState.ts`:

```js
t = {
  addtlConsent: "1~"
  cmpId: 6
  cmpStatus: "loaded"
  cmpVersion: 1
  eventStatus: "tcloaded"
  gdprApplies: false /* ← I don’t like the look of this… */
  listenerId: null
  tcfPolicyVersion: undefined
}
```

sndrs (approved):

Issue #239, opened 2020-08-28
RFC - helpers for common consents


sndrs:
After reviewing some TCFv2 PRs and seeing how the CMP is being used, would something like this be useful?

```js
import { getConsentFor, onConsentChange } from '@guardian/consent-management-platform`

onConsentChange(consent => {
    const ga = getConsentFor('google-analytics', consent); // true
    const comscore = getConsentFor('comscore', consent); // false

    // throws error
    const eowifnwoeifjoweinf = getConsentFor('eowifnwoeifjoweinf', consent); 

    // you can still use the consent state as now
    const complexConsentCondition = myComplexConsentTask(consent);
})
```

_EDIT 01.09.2020: Updated following https://github.com/guardian/consent-management-platform/issues/239#issuecomment-682471839_

Also, would this actually fit the consent models currently in operation?

e.g. https://github.com/guardian/support-frontend/pull/2650

GHaberis:
This would save us having to hold a reference to the vendor IDs in our code which we're currently doing so we can interrogate  `state.tcfv2.vendorConsents` in our `onConsentChange` callback, I imagine others are doing this too, and probably in a variety of ways. 

I like that this solution provides an interface for interrogating the `state` data object, so we don't have to, this means users of the library wouldn't necessarily need to know what `state` looks like, and if that `state` model were to change in the future users of this library wouldn't need to restructure their code to deal with it, the only changes necessary would be in the library as the interface would remain the same.

sndrs:
> This would save us having to hold a reference to the vendor IDs in our code

this feels brittle to me, def something we'd be better abstracting away. if you had cause to do something more complex you can always use the actual `__tcfapi` or `__uspapi`. but probably no one will need to do that.

sndrs:
actually maybe it would look more like

```js
import { getConsentFor, onConsentChange } from '@guardian/consent-management-platform`

onConsentChange(consent => {
    getConsentFor('ga', consent).then(ga => {
        // ga = true
    })

    getConsentFor('comscore', consent).then(comscore => {
        // comscore = false
    })

    getConsentFor('eowifnwoeifjoweinf') // rejects
})
```

ioanna0:
I like this refactor as it will simplify the consent management of the vendors on the listener side and encapsulate both their ids and the consent logic into the CMP library in one place. It will also be a lot easier to see the full list of vendors been used as a whole in the guardian without navigating all the repos.
[...]

PR #276, opened 2020-09-16
Add getConsentFor

sndrs:
refs #239

Creates a new module to facilitate the consent logic

Pending
- [x] Get all sourcepoint ids from all the repos


src/getConsentFor.ts
```diff
@@ -0,0 +1,34 @@
...
+export const getConsentFor = (
+	vendor: keyof typeof VendorIDs,
+	consent: ConsentState,
+): boolean => {
+	const sourcepointId = VendorIDs[vendor];
+	if (typeof sourcepointId === 'undefined') {
+		throw new Error(`Vendor '${vendor}' not found`);
+	}
+	const vendorConsent: boolean | undefined =
+		consent.tcfv2?.vendorConsents[sourcepointId];
```
sndrs, line 29: do we need to handle CCPA here too?

src/getConsentFor.ts
```diff
@@ -0,0 +1,34 @@
...
+	'a9' = '5edf9a821dc4e95986b66df4',
+}
+
+export const getConsentFor = (
+	vendor: keyof typeof VendorIDs,
+	consent: ConsentState,
+): boolean => {
+	const sourcepointId = VendorIDs[vendor];
+	if (typeof sourcepointId === 'undefined') {
+		throw new Error(`Vendor '${vendor}' not found`);
```
sndrs, line 26:
```suggestion
		throw new Error(`Vendor '${vendor}' not found. If it should be added, raise an issue at https://git.io/JUzVL`);
```

src/getConsentFor.ts
```diff
@@ -0,0 +1,34 @@
...
+	}
+	const vendorConsent: boolean | undefined =
+		consent.tcfv2?.vendorConsents[sourcepointId];
+	if (typeof vendorConsent === 'undefined') {
+		throw new Error(`No consent returned from SP for vendor: '${vendor}'`);
```
sndrs, line 31: what conditions could make this undefined? wondering if we need to return `false` in this instance, or offer a suggestion of something?

sndrs (commented):
be really good to update the readme too

src/getConsentFor.ts
ioanna0: One condition would be if the vendor is not setup properly inside Sourcepoint, or any other issue fetching the consent from sourcepoint. guess we could change it to false, with a warning instead

src/getConsentFor.ts
ioanna0: done 👍

README.md
```diff
@@ -197,6 +200,46 @@ onConsentChange(({ tcfv2, ccpa }) => {
...
+
+returns: `boolean`
+
+A method to get the consent for each particular custom vendor from Sourcepoint
+
+params:
+
+#### `vendor`: string representing the key of the Vendor Id
+    Check VendorIds (inside `src/types/index.ts`) to see if the vendor you are trying to use is already configured.
+    If not please raise an issue at https://git.io/JUzVL
```
mxdvl, line 219:
```suggestion
Check [VendorIds][] (inside `src/types/index.ts`) to see if the vendor you are trying to use is already configured.
If not please raise an issue at https://git.io/JUzVL

[VendorIds]: https://github.com/guardian/consent-management-platform/blob/9de2219cf2499871b3e2f11b4be42acad7ac6f96/src/types/index.ts#L38
```

src/getConsentFor.test.js
```diff
@@ -0,0 +1,44 @@
...
+	expect(getConsentFor('google-analytics', tcfv2ConsentFoundFalse)).toBeFalsy();
+});
+
+it('returns true ccpa consent if true', () => {
+	expect(getConsentFor('google-analytics', ccpaWithConsent)).toBeTruthy();
+});
+
+it('returns false ccpa consent if false', () => {
+	expect(getConsentFor('google-analytics', ccpaWithoutConsent)).toBeFalsy();
+});
```
mxdvl, line 44: May be worth using test tables here again?

mxdvl (approved):

README.md
```diff
@@ -197,6 +200,46 @@ onConsentChange(({ tcfv2, ccpa }) => {
...
+
+#### `consent`: the consent object been returned from `onConsentChange` callback
+
+
+### Example
```
sndrs, line 224:
```suggestion
#### Example
```

src/types/index.ts
```diff
@@ -30,3 +34,31 @@ export interface VendorConsents {
 		};
 	};
 }
+
+export enum VendorIDs {
```
sndrs, line 38: is this a type definition? it _is_ the actual data – i wonder if it should live in `getConsentFor.ts`?

README.md
```diff
@@ -197,6 +200,46 @@ onConsentChange(({ tcfv2, ccpa }) => {
 });
 ```
 
+## Get consent for particular vendor
```
sndrs, line 203: maybe this section can be collapsed into `Using Consent`? so it discusses both `getConsentFor` and `onConsentChange`?

src/types/index.ts
ioanna0: yeah is the actual data, i moved it inside index.ts because eslint was complaining about a circular dependency. open to suggestion on how to move it back to  `getConsentFor.ts` without that error :)

src/getConsentFor.test.js
```diff
@@ -0,0 +1,37 @@
...
+
+test.each([
+	['tcfv2', false, 'google-analytics', tcfv2ConsentNotFound],
+	['tcfv2', true, 'google-analytics', tcfv2ConsentFoundTrue],
+	['tcfv2', false, 'google-analytics', tcfv2ConsentFoundFalse],
+	['ccpa', true, 'google-analytics', ccpaWithConsent],
+	['ccpa', false, 'google-analytics', ccpaWithoutConsent],
+])(
+	`In %s mode, returns %s, for vendor %s`,
```
mxdvl, line 33: 👍

mxdvl (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@guardian/consent-management-platform",
-	"version": "6.0.0",
+	"version": "6.2.0-0",
```
sndrs, line 3: i don't htink this needs to be in here, `np` should work it out?
```suggestion
	"version": "6.0.0",
```

README.md
```diff
@@ -198,6 +199,46 @@ onConsentChange(({ tcfv2, ccpa }) => {
...
+    onConsentChange(consent => {
+        const ga = getConsentFor('google-analytics', consent); // true
+        const comscore = getConsentFor('comscore', consent); // false
+
+        // throws error
+        const eowifnwoeifjoweinf = getConsentFor('eowifnwoeifjoweinf', consent);
+
+        // you can still use the consent state for a more complicated logic
+        const complexConsentCondition = myComplexConsentTask(consent);
+    })
```
sndrs, line 237:
```suggestion
import { getConsentFor, onConsentChange } from '@guardian/consent-management-platform`

onConsentChange(consent => {
    const ga = getConsentFor('google-analytics', consent); // true
    const comscore = getConsentFor('comscore', consent); // false

    // throws error
    const eowifnwoeifjoweinf = getConsentFor('eowifnwoeifjoweinf', consent);

    // you can still use the consent state for a more complicated logic
    const complexConsentCondition = myComplexConsentTask(consent);
})
```

src/types/index.ts
sndrs: ah right, ok. how about this?

in `types/window.d.ts`, use `typeof` e.g. 

```ts
import { getConsentFor } from '../getConsentFor';
import { onConsentChange } from '../onConsentChange';
// ...
onConsentChange?: typeof onConsentChange;
getConsentFor?: typeof getConsentFor;
```

then you should be able to remove `OnConsentChange` and `GetConsentFor` types from `types/index.ts` entirely (also makes it safer, don't have to maintain the definitions).

then i think you could move the `VendorIDs` to `src/getConsentFor.ts`

package.json
ioanna0: will it include those changes into `6.0.0` once we merge them into main branch?

ioanna0: i was thinking to make another release for these with the latest changes, is it needed?

src/types/index.ts
ioanna0: much simpler ! 💯 done

README.md
```diff
@@ -198,6 +199,44 @@ onConsentChange(({ tcfv2, ccpa }) => {
...
+
+params:
+
+#### `vendor`: string representing the key of the Vendor Id
+Check [VendorIds][] (inside `src/getConsentFor.ts`) to see if the vendor you are trying to use is already configured.
+If not please raise an issue at https://git.io/JUzVL
+
+[VendorIds]: https://github.com/guardian/consent-management-platform/blob/b243844e2ec2e1c8b0b6849c248c308347de2cd9/src/getConsentFor.ts#L3
+
+#### `consent`: the consent object been returned from `onConsentChange` callback
```
sndrs, line 220:
```suggestion
#### `vendor`

type: string

See the list of [`VendorIDs `][] for options. If vendor you need is missing, please [raise an issue](https://git.io/JUzVL) (or a PR!). 

[VendorIDs]: https://github.com/guardian/consent-management-platform/blob/main/src/getConsentFor.ts

#### `consent`

type: `object` 

The consent object returned from `onConsentChange` callback.
```

PR #286, opened 2020-09-23
chore: update TCFv2_banner_amp.css

mxdvl:
## What does this change?

Some possible errors on large screens that use AMP have been fixed.

## How to test

Open an AMP page on a desktop device and see how the consent banner behaves.

Some users may not have support for `vh`. We don’t have a solution for them.

## Images

image


mxdvl:
Do you approve, @zeek01 ?

Let’s merge once we confirmed it works as expected.

sourcepoint-ui/TCFv2_banner_amp.css
```diff
@@ -409,7 +409,7 @@ https://github.com/guardian/consent-management-platform/tree/main/sourcepoint-ui
     
     .gu-content {
         border-top: 1px solid rgb(80, 105, 145) !important;
-        max-height: calc(51vh) !important;
+        max-height: calc(100vh) !important; /* See above */
```
sndrs, line 412: how does this behave where [`vh` is not supported](https://caniuse.com/?search=vh)?

mxdvl: Oh, that’s a good point. I had assumed support was more widespread.

In this case this only applies to AMP, so support is only relevant in mobile browsers. However, we are using similar rules in the main file to prevent from covering the whole page.

sndrs: i think it will still not work in older opera mobile, which is 0.28% of our readers for the last month:

mxdvl: AMP uses `vh` [in its codebase][1] already… How can we test this further?

[1]: https://github.com/ampproject/amphtml/blob/ab581081c76851f9e31e1558c9d06f8b4b4fbc52/extensions/amp-story/1.0/amp-story-consent.css

sourcepoint-ui/TCFv2_banner_amp.css
```diff
@@ -123,6 +123,7 @@ https://github.com/guardian/consent-management-platform/tree/main/sourcepoint-ui
 
 .gu-content {
     border-top: 1px solid rgb(80, 105, 145) !important;
+    max-height: calc(568px * 0.6) !important; /* Opera mini does not support viewport units. Calculates 60% of the height of an iPhone SE viewport */
```
mxdvl, line 126: I think you can do the calculation yourself and get rid of `calc`: `340px`.
Is this iPhone SE referring to 1<sup>st</sup> gen?

zeek01 (approved):
Looks good

PR #295, opened 2020-10-01
Add pageViewId and cmpInitTimeUtc to PubData

janua:
## What does this change?

This updates the `PubData` type to accept `pageViewId` and `cmpInitTimeUtc`.

We will use `pageViewId` to resolve the `browserId` for the consent signal, and phase out having `browserId` in this type.

We have already been relying in the server side timestamp from Sourcepoint when joining (which in fairness seems pretty accurate without much drift from most clients); however this still should not be relied upon IMO; so I have taken this opportunity to add `cmpInitTimeUtc`.

Maybe not the best name for `cmpInitTimeUtc`; was thinking perhaps of `cmpInitTimestampMillisUtc` as it is an Epoch timestamp in milliseconds.

To @sndrs point around `PubData` prescribing the keys within the object; I haven't changed this to a general object for now as I don't think there will be much movement in this type after this; if we find we are changing it a lot then we should move it to a more general type that accepts any keys.

We can monitor the presense of these new keys in the Sourcepoint logs.

@guardian/commercial-dev


src/types/index.ts
```diff
@@ -7,6 +7,8 @@ export interface ConsentState {
 }
 export interface PubData {
 	browserId?: string;
+	pageViewId?: string;
+	cmpInitTimeUtc?: number;
```
sndrs, line 11: is there a more accurate type you could use here?

janua: @sndrs What were you imaginging? Such as a DateTime? It's essentially a long

sndrs: yeah true, i guess it's just an infinite number of milliseconds?

README.md
```diff
@@ -82,7 +82,14 @@ Pass additional parameters for for reporting. Optional.
 #### Example
 
 ```js
-cmp.init({ pubData: { browserId: 'gow59fnwohwmshz' }, isInUsa: false });
+cmp.init({
+    pubData: {
+        browserId: 'gow59fnwohwmshz',
+        pageViewId: 'jkao3u2kcbaqk',
+        cmpInitTimeUtc: 1601511014537
```
sndrs, line 89: where would this number come from in practice? `new Date().getTime()`?

README.md
```diff
@@ -82,7 +82,14 @@ Pass additional parameters for for reporting. Optional.
 #### Example
 
 ```js
-cmp.init({ pubData: { browserId: 'gow59fnwohwmshz' }, isInUsa: false });
+cmp.init({
+    pubData: {
+        browserId: 'gow59fnwohwmshz',
```
sndrs, line 87: it would be useful to add a description of where these values could be acquired

janua: @sndrs Description added

sndrs (approved):

PR #303, opened 2020-10-13
feat(api)!: Add Australian framework `AUS`

mxdvl:
## What does this change?

This changes introduces a new “framework” for Australia—called **AUS**—based off Sourcepoint’s custom CCPA implementation. The key changes are as follows:

- **BREAKING:** new interface for initialising the CMP, takes `country` instead of `isInUsa`
- ~~Extension of CCPA’s API to include `getCustomVendorRejects`~~
- New API for APF (85% code copied from CCPA)
- New `Framework` type, accepting `tcfv2`, `ccpa` or `aus`
- New privacy manager ID for AUS and new property in Sourcepoint
- Update of Svelte app to support AUS

Users in Australia get an opt-out framework, similar to CCPA.

N/A

## Images

AUS privacy manager


src/index.ts
```diff
@@ -24,10 +26,10 @@ const initialised = new Promise((resolve) => {
 
 function init({
 	pubData,
-	isInUsa,
+	framework,
```
sndrs, line 29: this feels like it asks consumers of CMP to know a lot about decisions the business has made, and how the CMP works internally i.e. what frameworks are there, and which one applies where?

i _think_ we only need to know where the user is? 

so rather than `isInUsa` we could require something like `location: 'US' | 'AU' | 'RoW'`?

mxdvl: Yes, I agree!

I was also thinking we should accept any country code, rather than `RoW`. This would mean, for example, that a `NZ` or `CA` user could be put in a different and specific framework without any of the code consuming `@guardian/consent-management-platform` to change at all.

sndrs: oh yeah great idea

README.md
```diff
@@ -14,21 +14,21 @@ and TCFv2 to everyone else.
...
-  * [`cmp.showPrivacyManager()`](#cmpshowprivacymanager)
-- [Using Consent](#using-consent)
-  * [`onConsentChange(callback)`](#onconsentchangecallback)
-  * [`getConsentFor(vendor, consentState)`](#getconsentforvendor-consentstate)
-- [Disabling Consent](#disabling-consent)
-  * [`cmp.__disable()`](#cmp__disable)
-  * [`cmp.__enable()`](#cmp__enable)
-  * [`cmp.__isDisabled()`](#cmp__isdisabled)
-  * [Manually](#manually)
-- [Development](#development)
```
mxdvl, line 31: What’s this about?

README.md
```diff
@@ -66,11 +66,12 @@ Adds the relevent privacy framework to the page. It must be called to enable
...
 
-#### `options.isInUsa`
+#### `options.country`
 
-type: `boolean`
+type: `string`
+values: any 2-letter country code, e.g. `GB`, `US`, `AU`, …
 
-Declare whether your user is in the USA or not. Required – *throws an error if
+Declare which country yout user is in. Required – *throws an error if
```
mxdvl, line 74:
```suggestion
Declare which country your user is in. Required – *throws an error if
```

src/onConsentChange.ts
```diff
@@ -19,7 +20,10 @@ const invokeCallback = (callback: CallbackQueueItem, state: ConsentState) => {
 
 const getConsentState: () => Promise<ConsentState> = async () => {
 	if (window.__uspapi) {
-		// in USA - https://git.io/JUOdq
+		// in USA or AUS - https://git.io/JUOdq
+		if (window.guCmpHotFix.framework === 'aus')
```
mxdvl, line 24: Any thoughts on this?

sndrs: should we introduced a singleton that stores the active framework?

mxdvl: That would make sense!

sndrs: something like 

```ts
// currentFramework.ts
let currentFramework: Framework?;

export const getCurrentFramework= () => currentFramework;
export const setCurrentFramework = (framework: Framework) => currentFramework = framework;
```
then we can import it anywhere in the CMP, and set it during `init`

what do you think?

src/index.ts
```diff
@@ -38,15 +41,31 @@ function init({
...
+		case 'tcfv2':
+		default:
+			CMP = TCFv2;
+			break;
+		case 'ccpa':
+			CMP = CCPA;
+			break;
+		case 'aus':
+			CMP = AUS;
+			break;
```
sndrs, line 64: not a big deal, it just took me a while to work out what was happening here
```suggestion
		case 'ccpa':
			CMP = CCPA;
			break;
		case 'aus':
			CMP = AUS;
			break;
		case 'tcfv2':
		default: // default is also 'tcfv2'
			CMP = TCFv2;
			break;
```

src/onConsentChange.ts
sndrs: discussed irl

src/index.ts
mxdvl: I checked and `default:` does not have to be at the bottom. If the convention makes it easier to parse, happy to apply your suggestion.

sndrs: yeah i htink it will work just fine, i was just a bit lost trying to work it out, that's all. not a blocker :)

sndrs (approved):
nice!

PR #306, opened 2020-10-14
tests(e2e): Fix integrations tests

mxdvl:
## What does this change?

It seems that the iframe IDs can change, possibly when a re-permission is done.

This PR updates these IDs to the latest versions, and fixes integration tests.

A similar problem was seen in https://github.com/guardian/dotcom-rendering/pull/1969

## How to test

- `yarn test:e2e`
- Checks should now pass!


sndrs:
coudl this make use of the attribute selector to find it? say `iframe[id*="sp_message_iframe"]`?

mxdvl:
> coudl this make use of the attribute selector to find it? say `iframe[id*="sp_message_iframe"]`?

We could do, but it could also overlap with the privacy manager… In some ways it seems okay that the CMP tests are capturing these changes.

ioanna0:
could we ask SP to add a uniquely identifier for us?

sndrs (approved):

mxdvl:
@ioanna0 I think the id changes whenever the vendor list is updated. We can always ask, but it’s unlikely to work with their system.

PR #318, opened 2020-10-20
fix: use object type rather than enum to store countries

sndrs:
## What does this change?

uses object type for countries

## Why?

so we can get a union type of the value for input


mxdvl (approved):
This seems sound!

PR #320, opened 2020-10-21
fix: accept `string` as a country (for now)

sndrs:
## What does this change?

- allows `string` types for country

## Why?

- DCR `countryCode` type is string, so this is a quick fix for Aus release

## Then...
- we could pull [the geo stuff out of DCR](https://github.com/guardian/dotcom-rendering/blob/main/src/web/lib/getCountryCode.ts) into https://github.com/guardian/libs and set the correct ISO type in there (cc @gtrufitt @liywjl)
- we could then revert this PR

## Then 2...
- that also opens up the possibility of not needing to pass location in, we can use the lib ourselves

### DIFF
--- a/src/types/countries.ts
+++ b/src/types/countries.ts
@@ -253,4 +253,5 @@ type Countries = {
 	Zimbabwe: 'ZW';
 };
 
-export type Country = Countries[keyof Countries];
+// temporarily accept any string to integrate into DCR
+export type Country = Countries[keyof Countries] | string;


gtrufitt (approved):
+1 geo lib is a good shout

sndrs:
also, something weird is happening with the README toc generation, ignore for now i'll look into it

README.md
```diff
@@ -12,23 +12,25 @@ and TCFv2 to everyone else.
 
 ## Table of Contents
 
+<!-- this TOC is auto generated – don't edit it :) -->
+
 <!-- toc -->
 
--   [Installation](#installation)
```
sndrs, line 17: this shouldn't be happening – its auto generated, i'll look into it in a seperate PR

src/types/countries.ts
```diff
@@ -1,256 +1,260 @@
...
-	WallisAndFutuna: 'WF';
-	WesternSahara: 'EH';
-	Yemen: 'YE';
-	Zambia: 'ZM';
-	Zimbabwe: 'ZW';
-};
+// temporarily disable this to integrate into DCR
 
-export type Country = Countries[keyof Countries];
+export type Country = string;
```
mxdvl, line 6: I think it makes more sense to just do a string union?

```suggestion
export type Country = Countries[keyof Countries] | string;
```

mxdvl (changes requested):

src/types/countries.ts
sndrs: that just ends up being `string`, because all countries are strings too

sndrs: `'AU' | string` will accept any string :)

mxdvl: I just meant in terms of the amount of code change, so that in the future, reverting is only removing the union.

README.md
mxdvl: I think it was changed when I edited them, too. Maybe an odd prettier setting?

mxdvl (approved):
Ignore my qualms about unions.

Ship it if needed!

src/types/countries.ts
sndrs: actually i think you're right, it's a much smaller change

PR #321, opened 2020-10-21
fix: add AUS getConsentFor

mxdvl:
## What does this change?

The `getConsentFor` method did not support the AUS framework.

## Why?

This needed to be updated before it could be used in production.

## Potential Further work

It might be worth changing the shape of the AUS consent state to make it easier to handle.

### DIFF
--- a/src/getConsentFor.ts
+++ b/src/getConsentFor.ts
@@ -47,7 +47,8 @@ export const getConsentFor = (
 	if (consent.aus) {
 		if (typeof consent.aus.rejectedVendors === 'undefined') return true;
 		const rejected = consent.aus.rejectedVendors.filter(
-			(rejectedVendor) => rejectedVendor.[`_id`] === sourcepointId,
+			// eslint-disable-next-line no-underscore-dangle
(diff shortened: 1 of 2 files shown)


sndrs (approved):

src/getConsentFor.ts
```diff
@@ -44,6 +44,15 @@ export const getConsentFor = (
 		return !consent.ccpa.doNotSell;
 	}
 
+	if (consent.aus) {
+		if (typeof consent.aus.rejectedVendors === 'undefined') return true;
+		const rejected = consent.aus.rejectedVendors.filter(
+			// eslint-disable-next-line no-underscore-dangle
+			(rejectedVendor) => rejectedVendor._id === sourcepointId,
```
ioanna0, line 51:
```suggestion
			(rejectedVendor) => rejectedVendor.[`_id`] === sourcepointId,
```

mxdvl: I’m afraid that’s not going to work with prettier 😞

PR #322, opened 2020-10-21
feat: AUS handle old API params

mxdvl:
## What does this change?

Allow calling the API with `cmp.init({ isInUsa })`.

## Why?

Because we can never fully synchronise releases in DCR & frontend, we always need to support old API interfaces.


sndrs (approved):

sndrs (approved):
lovely

PR #323, opened 2020-10-21
perf: Expire cookies far into the past

mxdvl:
## What does this change?

Set to a relevant date in The Guardian’s history.


mxdvl:
Looking into it, it doesn’t seem to solve the problem of rogue cookies on the test page…

mxdvl:
This is because the preferences are not saved on the domain the test page is served from. Javascript does not have access to these cookies.

cc. @sndrs

PR #324, opened 2020-10-23
fix: add new types back to window object

sndrs:
## What does this change?

restores old types to init interface

## Why?

missed in #322


mxdvl (approved):
Good shout!

PR #333, opened 2020-10-29
feat: Add Ipsos Mori vendor ID

nitro-marky:
## What does this change?

Adds the Ipsos Mori vendor ID


src/getConsentFor.ts
```diff
@@ -16,6 +16,7 @@ enum VendorIDs {
 	'googletag' = '5f1aada6b8e05c306c0597d7',
 	'ias' = '5e7ced57b8e05c485246ccf3',
 	'inizio' = '5e37fc3e56a5e6615502f9c9',
+	'ipsos' = '5f745ab96f3aae0163740409',
```
sndrs, line 19: can you add this to the readme too? just so everything is in sync

sndrs (changes requested):
i think you'll need to update the PR title to `feat:` too, otherwise you won't trigger a release – see https://github.com/guardian/consent-management-platform/blob/main/docs/01-development-instructions.md#releasing

sndrs (approved):
looks good

PR #334, opened 2020-10-30
chore: use @guardian eslint configs

sndrs:
## What does this change?

- counterpart to #317 (and dependant on it)
- uses global guardian configs

## Why?

- consistency and good practise


cypress/integration/sourcepoint-tcfv2.spec.js
```diff
@@ -57,7 +55,7 @@ skipOn(Cypress.env('CI') === 'true', () => {
 				.find(`button[title="${buttonTitle}"]`)
 				.click();
 
-			// eslint-disable-next-line cypress/no-unnecessary-waiting
+			// eslint-disable-next-line cypress/no-unnecessary-waiting -- should we do this?
```
mxdvl, line 58: I think we can wait for a cookie to be present… let’s look at how it’s done in CCPA and raise another PR.

src/aus/sourcepoint.test.js
```diff
@@ -44,7 +43,7 @@ describe('Sourcepoint CCPA', () => {
 	it('points at a real file', (done) => {
 		init();
 		const src = document
-			?.getElementById('sourcepoint-aus-lib')
+			.getElementById('sourcepoint-aus-lib')
```
mxdvl, line 46: Why is this gone?

src/lib/domain.ts
```diff
@@ -2,6 +2,6 @@ let isGuardian: boolean | undefined;
 
 export const isGuardianDomain = (): boolean => {
 	if (typeof isGuardian === 'undefined')
-		isGuardian = /\.theguardian\.com$/.test(window.location.host);
+		isGuardian = window.location.host.endsWith('.theguardian.com');
```
mxdvl, line 5: ’tis such Poetry…

src/onConsentChange.ts
```diff
@@ -51,9 +49,7 @@ export const onConsentChange: (fn: Callback) => void = (callBack) => {
 	// if consentState is already available, invoke callback immediately
 	getConsentState()
 		.then((consentState) => {
-			if (consentState) {
-				invokeCallback(newCallback, consentState);
-			}
+			invokeCallback(newCallback, consentState);
```
mxdvl, line 52: So `consentState` is always defined?

src/types/tcfv2/TCData.ts
```diff
@@ -88,15 +88,6 @@ export interface TCData {
 			consents: TCFv2ConsentList;
 			legitimateInterests: TCFv2ConsentList;
 		};
-		restrictions: {
-			[key: string]: {
-				/**
-				 * 0 - Not Allowed
-				 * 1 - Require Consents
-				 * 2 - Require Legitimate Interest
-				 */
```
mxdvl, line 97: We’ve lost this comment! Worth turning into a type?

tsconfig.json
```diff
@@ -10,5 +10,5 @@
     "strict": true,
     "target": "ES2020"
   },
-  "exclude": ["node_modules", "dist", "coverage", "cypress", ".*"]
+  "include": ["*.js", ".*.js", "src"]
```
mxdvl, line 13: What about `node_modules`?

mxdvl (approved):
Few comments, but generally 👍

tsconfig.json
sndrs: it is excluded by default

src/lib/domain.ts
sndrs: yeah auto-fixed too!

src/types/tcfv2/TCData.ts
sndrs: oh no that was an oversight, restoring it

src/aus/sourcepoint.test.js
sndrs: i guess it was not necesary, `document` must be defined in jest types. it was removed by a fixable rule

src/onConsentChange.ts
sndrs: that's what typescript thinks. i wonder if we should update the type so it can be `undefined`?

mxdvl: In TS we trust.

PR #342, opened 2020-11-02
feat!: Use getLocale

sndrs:
## What does this change?

- uses `getLocale` from https://github.com/guardian/libs instead of requiring it to be passed to `init` 

## Why?

- cleaner API
- its the same technique host apps would use, but just handled internally


mxdvl (approved):
This is so much cleaner, we handle all locales: 🇬🇧 🇺🇸 🇦🇺 🇪🇺 !

ioanna0 (approved):

src/index.ts
```diff
@@ -37,42 +34,34 @@ const init: InitCMP = ({
...
+			case 'tcfv2':
+			default:
+				// default is also 'tcfv2'
+				CMP = TCFv2;
+				break;
+		}
+
+		setCurrentFramework(framework);
+
+		CMP.init(pubData ?? {});
```
mxdvl, line 60: Is it even worth passing an empty object?

```suggestion
		CMP.init(pubData);
```

mxdvl (commented):
Simple and elegant.

I think there should we should decide on what kind of error we throw if old parameters are passed. Technically, we could simply say “`isInUsa` and `country` are ignored, assumed locale is `GB`, defined from @guardian/libs”.

There should also be tests that capture this behaviour. I know that some projects, such as `support-frontend`, which are relying on a different approach to the `@guardian/libs` to set the locale, might find it difficult to work with our new interface.

src/types/index.ts
```diff
@@ -9,7 +9,7 @@ export type InitCMP = (arg0: {
 	pubData?: PubData;
 	country?: Country;
```
mxdvl, line 10:
```suggestion
```

mxdvl (commented):
We also need to ensure that we mock the result of `getLocale` for the test page.

It might be as simple as setting the `gu.geolocation` value in the Svelte App.

mxdvl:
Yea, it might be an improvement, but not urgent in any way an no consumer has requested this change so far.

PR #349, opened 2020-11-09
fix: only test against advertising category

mxdvl:
## What does this change?

In AUS, Instead of testing agains specific custom vendors, check for the “Advertising” category.

## Why?

This decision has been clarified as being the expected behaviour.

### DIFF
--- a/src/getConsentFor.ts
+++ b/src/getConsentFor.ts
@@ -46,10 +46,11 @@ export const getConsentFor = (
 	}
 
 	if (consent.aus) {
-		if (typeof consent.aus.rejectedVendors === 'undefined') return true;
-		const rejected = consent.aus.rejectedVendors.filter(
+		if (typeof consent.aus.rejectedCategories === 'undefined') return true;
+		const advertising = '5f859c3420e4ec3e476c7006';
+		const rejected = consent.aus.rejectedCategories.filter(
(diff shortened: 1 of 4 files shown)


ioanna0 (approved):

ioanna0:
Just needs a unit test fix

mxdvl:
Thanks @ioanna0, fixed now.

I’ve made the iframe selector more generic, as its ID changes when the custom vendor list changes.

src/getConsentFor.ts
```diff
@@ -46,10 +46,11 @@ export const getConsentFor = (
 	}
 
 	if (consent.aus) {
-		if (typeof consent.aus.rejectedVendors === 'undefined') return true;
-		const rejected = consent.aus.rejectedVendors.filter(
+		if (typeof consent.aus.rejectedCategories === 'undefined') return true;
+		const advertising = '5f859c3420e4ec3e476c7006';
```
ioanna0, line 50: sorry for the post-release comment. just had a small refactoring idea, if we can move the `advertising` const as a new key into the `VendorIDs` and then we can access it directly with `VendorIDs.advertising

mxdvl: I did think that too, actually. We should call it 'aus-advertising'

mxdvl: #353

PR #354, opened 2020-11-13
fix: wait for sourcepoint library to load in aus

sndrs:
## What does this change?

delays calls to `uspapi` until the sourcepoint library has loaded in aus

## Why?

sourcepoint's custom `'getCustomVendorRejects'` extension to `__uspapi` doesn't implement the queue that the standard `__uspapi` options are required to. 

they suggested the following:

> wait until we have determined the siteID before calling __uspapi('getCustomVendorRejects');

this is an attempt but it doesn't seem to work...

### DIFF
--- a/cypress/utils.js
+++ b/cypress/utils.js
@@ -1,0 +1,9 @@
+export const loadPage = (url) => {
+	it(`should load the page: ${url}`, () => {
+		cy.visit(url);
+
+		// undocumented fix for cookies from non-localhost domains
+		// https://github.com/cypress-io/cypress/issues/408#issuecomment-643281127
+		cy.clearCookies({ domain: null });
+	});
+};
--- a/cypress/utils.js
+++ b/cypress/utils.js
@@ -1,9 +1,9 @@
 export const loadPage = (url) => {
 	it(`should load the page: ${url}`, () => {
-		cy.visit(url);
-
 		// undocumented fix for cookies from non-localhost domains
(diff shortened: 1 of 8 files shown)


mxdvl:
This looks like it’s timing out.

src/aus/sourcepoint.ts
```diff
@@ -47,6 +52,14 @@ export const init = (pubData = {}): void => {
 			events: {
 				onConsentReady() {
 					mark('cmp-aus-got-consent');
+
+					// the 'getCustomVendorRejects' option of SP's implementation of __uspapi
+					// is a custom extension. It hits SP's servers, but unlike the rest of the
+					// __uspapi, it doesn't implement a queue.
+					// the only way we can be sure it has become available is to wait for a
+					// SP event to fire, so we resolve this now so we can be sure its available elsewhere
+					void resolveLoaded();
```
ioanna0, line 61: 👍

mxdvl (approved):
💣 ¡Death To All Cookies! 🍪

cypress/utils.js
```diff
@@ -1,9 +1,9 @@
 export const loadPage = (url) => {
 	it(`should load the page: ${url}`, () => {
-		cy.visit(url);
-
 		// undocumented fix for cookies from non-localhost domains
 		// https://github.com/cypress-io/cypress/issues/408#issuecomment-643281127
 		cy.clearCookies({ domain: null });
+
+		cy.visit(url);
```
sndrs, line 7: nice

PR #360, opened 2020-11-16
fix: await __uspapi

mxdvl:
## What does this change?

More robust checks for when SP library has loaded. Any callback events now triggers the `getCustomVendorRejects` call. Follow-up on #354. 

- Renamed promise to `sourcepointLibraryLoaded`

## Why?

`onConsentReady` does not fire when a user gets the banner show and clicks “Continue”. Using the `willShowPrivacyMessage` is more robust, and always fires once the SP library is loaded.

### DIFF
--- a/src/aus/api.test.js
+++ b/src/aus/api.test.js
@@ -1,7 +1,7 @@
 import { getCustomVendorRejects } from './api';
 
 jest.mock('./sourcepoint', () => ({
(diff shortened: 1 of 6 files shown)


sndrs (approved):

mxdvl (commented):
I approve my own PR

(We've paired with @sndrs)

PR #362, opened 2020-11-18
fix: use CCPA getUSPData API in AUS

mxdvl:
## What does this change?

Just rely on good old `getUSPData` in AUS.

## Why?

We're not using the granularity of `getCustomVendorRejects`, and it's a half-baked solution that causes a plethora of errors everywhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,22 +14,22 @@ and TCFv2 to everyone else.
 
 <!-- toc -->
 
-- [Installation](#installation)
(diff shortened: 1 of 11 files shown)


sndrs (approved):
lgtm

ioanna0 (approved):
👍

mxdvl:
This change was conducted in pairing with @ashishpuliyel.

PR #363, opened 2020-11-18
fix: invert DNS logic

mxdvl:
## What does this change?

Do the right thing #362

## Why?

Consent is inverted otherwise.

### DIFF
--- a/src/aus/getConsentState.ts
+++ b/src/aus/getConsentState.ts
@@ -1,9 +1,11 @@
-import type { CustomVendorRejects } from '../types/aus';
-import { getCustomVendorRejects } from './api';
+import type { AUSConsentState } from '../types/aus';
+import { getUSPData } from './api';
 
 // get the current constent state using the official IAB method
-export const getConsentState: () => Promise<CustomVendorRejects> = async () => {
-	const customVendorRejects = await getCustomVendorRejects();
+export const getConsentState: () => Promise<AUSConsentState> = async () => {
+	const uspData = await getUSPData();
 
-	return customVendorRejects;
+	return {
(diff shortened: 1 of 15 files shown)


ashishpuliyel (approved):
Looks good to me

src/aus/getConsentState.ts
```diff
@@ -6,6 +6,6 @@ export const getConsentState: () => Promise<AUSConsentState> = async () => {
 	const uspData = await getUSPData();
 
 	return {
-		personalisedAdvertising: uspData.uspString.charAt(2) === 'Y',
+		personalisedAdvertising: uspData.uspString.charAt(2) !== 'Y',
```
ioanna0, line 9: wondering if we could have keep the same name as CCPA which is negative `doNotSell` . Not blocking of course

ioanna0 (approved):

src/aus/getConsentState.ts
mxdvl: The third character is either `Y`, `N` or `-`.
We want personalised advertising if it's either `N`, or `-`.

[As per the docs](https://github.com/InteractiveAdvertisingBureau/CCPA-reference-code):

> The usprivacy string as the format: `vnol` where
`v` = version (int)
`n` = Notice Given (char)
`o` = OptedOut (char)
`l` = Lspact (char)

mxdvl: I think it's better to have our code match our copy (ON / OFF &larr;&rarr; `true` / `false`)

mxdvl: I've clarified the logic in two steps with b86ee19.

PR #371, opened 2020-12-02
feat: add Sync methods

sndrs:
## What does this change?

adds 2 new methods to `cmp`:

- `willShowPrivacyMessageSync`
- `hasInitialised`

## Why?

an issue with Australian page skins means we need to these async options, in a situation where we know the CMP has kicked off


README.md
```diff
@@ -125,6 +133,26 @@ cmp.willShowPrivacyMessage()
     );
 ```
 
+### `cmp.willShowPrivacyMessageSync()`
+
+returns: `Boolean`
+
+Returns `true` if the CMP will show the initial privacy message (or is, or was), or `false` if not.
```
mxdvl, line 140:
```suggestion
Returns `true` if the CMP will show the initial privacy message, or `false` if not.
```

src/index.ts
```diff
@@ -57,40 +62,53 @@ const init: InitCMP = ({
...
+	activeCMP.init(pubData ?? {});
 	resolveInitialised();
 };
 
 const willShowPrivacyMessage: WillShowPrivacyMessage = () =>
-	initialised.then(() => CMP?.willShowPrivacyMessage() ?? false);
+	initialised.then(() => activeCMP?.willShowPrivacyMessage() ?? false);
+
+const willShowPrivacyMessageSync = () => {
+	if (_willShowPrivacyMessage !== void 0) {
```
mxdvl, line 87: Why not just use `undefined`? 

> The `void` operator is often used merely to obtain the `undefined` primitive value, usually using "`void(0)`" (which is equivalent to "`void 0`"). In these cases, the global variable [`undefined`][] can be used.
>
> _Source: [MDN: void](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void)_

[`undefined`]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined

```suggestion
	if (_willShowPrivacyMessage !== undefined) {
```

mxdvl (commented):
This seems like a sound addition.

We probably want to move the following line outside of `init` and into `initialised`:

```js
window.guCmpHotFix.initialised = true;
```

src/index.ts
sndrs: `undefined` used to be able to be redefined. that's prob less dangerous now, it's an old habit. but `typeof void 0 === 'undefined'` is _always_ true

README.md
sndrs: how come?

src/index.ts
mxdvl: as in, `'undefined'`, the string?

README.md
mxdvl: It’s the triple-“or” that’s confusing.

> `1` if `A` (or `B`, or `C`), or `0` if not.

Could we make it as below?

```suggestion
Returns `true` if the CMP has shown, is showing or will show the initial privacy message. Returns `false` otherwise.
```

I don’t think too strongly about this in any case.

src/index.test.js
```diff
@@ -125,6 +125,34 @@ describe('cmp.willShowPrivacyMessage', () => {
...
 
+describe('cmp.willShowPrivacyMessageSync', () => {
+	it('throws if CMP is not initialised', () => {
+		expect(() => cmp.willShowPrivacyMessageSync()).toThrow();
+	});
+
+	it('does not throw if CMP is initialised', () => {
+		cmp.init({ country: 'GB' });
+
+		cmp.willShowPrivacyMessage().then(() => {
```
mxdvl, line 136: It’s funny we’re still using the resolved state of `willShowPrivacyMessage` as a way to check the CMP has initialised.

Would it make sense for `init` to return a Promise?

mxdvl (approved):
LGTM.

As mentioned below, I still find it a little odd we’re relying on `cmp.willShowPrivacyMessage` as a way to check the CMP has initialised.

src/index.ts
ashishpuliyel: To be clear `void([any valid expression])` and returns undefined. So `void 0` (or `void 2121` or `void Math.PI`) always = `undefined`.
This is also why some libraries would wrap in a ```function(undefined) {
// all the code here safe in the knowledge that undefined is actually undefined
}()``` 

But this is a history lesson. Since undefined it's no longer writeable I vote accepting Max's suggestion. Simpler. It's not an idiom that you see much anymore and no need to confuse the young'ns.

src/index.test.js
ashishpuliyel: I think it's fine. Because everyone who cares is using onConsentChange. Init already returns a promise.
But if we wanted to I think the change would be this:

```
void frameworkCMP.willShowPrivacyMessage().then((willShowValue) => {
		_willShowPrivacyMessage = willShowValue;
		initComplete = true;
	}).then(resolveInitialised);
```

ashishpuliyel (approved):

src/index.test.js
mxdvl: `init` currently returns `void`, but I agree as this is internal use there's no need to really worry about it.

src/index.ts
sndrs: I seem to recall @janua had a plan to write a jquery-style `undefined.js` library once, where all the utilities hung off `undefined`instead

sndrs: so much for my 'early adopter' approach @mxdvl!

src/index.ts
```diff
@@ -53,44 +64,61 @@ const init: InitCMP = ({
...
 
 	setCurrentFramework(framework);
 
-	CMP.init(pubData ?? {});
+	frameworkCMP.init(pubData ?? {});
+
+	void frameworkCMP.willShowPrivacyMessage().then((willShowValue) => {
+		_willShowPrivacyMessage = willShowValue;
+		initComplete = true;
+	});
```
mxdvl, line 88: Is this `void` here to ensure the following expression is executed?

PR #372, opened 2020-12-02
docs: Small corrections to README

ripecosta:
## What does this change?
Does 3 small corrections to the README. 

## Why?
- `Expected parameters` was next to `options.isInUsa` instead of next to `options.pubData`, which didn't seem like the right place given the example below it.
- Docs stated that `consentState.tcfv2` would be undefined if user was outside the US or Australia when it's the other way around.
- Typo in a comment fixed while I was in the area.


mxdvl (approved):
👍

PR #378, opened 2020-12-09
feat: use @guardian/libs logger

mxdvl:
## What does this change?

Uses the new and shiny `log` from `@guardian/libs`.

## Why?

We can log more to the console, without polluting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "@babel/preset-typescript": "7.12.1",
     "@babel/runtime": "7.12.5",
     "@cypress/skip-test": "2.5.0",
+    "@guardian/libs": "1.6.1",
     "@guardian/eslint-config": "0.4.0",
     "@guardian/eslint-config-typescript": "0.4.1",
     "@guardian/prettier": "0.4.1",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -48,6 +48,7 @@
     "@babel/preset-typescript": "7.12.1",
     "@babel/runtime": "7.12.5",
     "@cypress/skip-test": "2.5.0",
+    "@guardian/libs": "1.6.1",
```
sndrs, line 51: you'll need to add this as a peerDep too, to make sure it's available when the CMP is installed. prob a good idea to add that as `^1.6.1`

package.json
```diff
@@ -48,7 +48,6 @@
     "@babel/preset-typescript": "7.12.1",
     "@babel/runtime": "7.12.5",
     "@cypress/skip-test": "2.5.0",
-    "@guardian/libs": "1.6.1",
```
sndrs: sorry max, you'll need it in here too for development purposes

sndrs (changes requested):

sndrs (approved):
nice!

Issue #382, opened 2020-12-16
[TCFv2] onConsentChange callbacks only trigger after user interacts with banner


ripecosta:
*Use case*: A fresh user (no cookies) accesses the website from a country where TCFv2 applies.

*Expected behaviour*: The `onConsentChange` triggers the callbacks with a non-consented state as soon as the CMP has finished initialising and before the user has interacted with the banner. (based on the [docs](https://www.npmjs.com/package/@guardian/consent-management-platform#onconsentchangecallback))

*Observed behaviour*: `onConsentChange` callbacks get triggered only after the user has interacted with the banner. Tested in manage-frontend's integration (not yet live) and confirmed in the [GitHub CMP test page](https://guardian.github.io/consent-management-platform/#tcfv2)

PR #383, opened 2020-12-16
fix: Invoke callbacks instantly in TCFv2

mxdvl:
## What does this change?

Returns a state with no consent as soon as possible, before a new user interacts with the banner.

## Why?

We don’t need to wait for SP events. Closing #382

### DIFF
--- a/src/aus/sourcepoint.ts
+++ b/src/aus/sourcepoint.ts
@@ -28,10 +28,6 @@ const properties = {
 export const init = (pubData = {}): void => {
 	stub();
 
-	// invoke callbacks ASAP in AUS
-	// TODO this causes an error
-	invokeCallbacks();
-
 	// make sure nothing else on the page has accidentally
 	// used the _sp_* name as well
 	if (window._sp_ccpa) {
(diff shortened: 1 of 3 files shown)


ripecosta:
I believe it's best to fix this bug by triggering the callbacks on the `onMessageReady` event (described [here](https://documentation.sourcepoint.com/web-implementation/web-implementation/sourcepoint-gdpr-and-tcf-v2-support/optional-callbacks-tcfv2)). This way you ensure that Sourcepoint is definitely ready when you trigger the callbacks and avoid adding unnecessary noise (no extra unnecessary callback calls). Let me know if you want to discuss this further.

mxdvl:
Thanks for rising, @ripecosta. [From Sourcepoint docs][1]:

> onConsentReady - this event fires when the consent object is ready. The consentUUID and `euconsent` values are sent to the callback. This does not apply to the TCF consent string. You should use the `addEventListener` command of the `__tcfapi` function and listen for the `tcloaded` event to obtain the user's consent string as soon as it available. More information can be found [here](https://documentation.sourcepoint.com/web-implementation/web-implementation/sourcepoint-gdpr-and-tcf-v2-support/using-__tcfapi-addeventlistener-and-removeeventlistener-commands).

[1]: https://documentation.sourcepoint.com/web-implementation/web-implementation/sourcepoint-gdpr-and-tcf-v2-support/optional-callbacks-tcfv2

ripecosta:
@mxdvl I didn't realise `invokeCallbacks` was already waiting for the CMP to be ready and checking if the consent changed. All good then!

ripecosta (approved):
👍
Thanks for addressing this so quickly.

PR #394, opened 2021-01-11
fix: update bg colour from 052862 to 052962

SiAdcock:
## What does this change?

Update instances of `#052862` to `#052962`

Millions of lines of code have been updated to meet the `.editorconfig`. Let me know if you'd prefer me to ignore `.editorconfig` for this change.

## Why?

CMP uses the colour `#052862` extensively. A [similar colour](https://theguardian.design/2a1e5182b/p/492a30-light-palette/t/66411c) is specified in Source: `#052962`. For consistency, we should converge on the colour in Source.

Please let me know if this is the right way to update these colours! 😅


mxdvl:
Thanks @SiAdcock, I think you can leave out the `.editorconfig` for this change so we can review it properly. We can pick this up later with @zeek01.

SiAdcock:
Thanks @mxdvl. I've disabled all the auto-formatting functionality in VS Code but it still insists on removing extraneous whitespaces 😬  Can you review by hiding whitespace changes in the GitHub UI?

Screenshot 2021-01-20 at 09 23 18

mxdvl (approved):
Good catch, thanks!

PR #396, opened 2021-01-14
fix: change type of version to string

ioanna0:
## What does this change?

Attempt to fix the below error when importing latest version to DCR which uses typescript:

```
node_modules/@guardian/consent-management-platform/dist/types/index.d.ts:12:21 - error TS2304: Cannot find name '__PACKAGE_VERSION__'.

12     version: typeof __PACKAGE_VERSION__;
```

## Why?
It should allow updating latest cmp version to DCR

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -11,7 +11,7 @@ export type CMP = {
 	willShowPrivacyMessageSync: () => boolean;
 	hasInitialised: () => boolean;
 	showPrivacyManager: () => void;
-	version: typeof __PACKAGE_VERSION__;
+	version: string;
 	__isDisabled: () => boolean;
 	__disable: () => void;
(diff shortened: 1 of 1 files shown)


mxdvl (commented):
Seems reasonable… although rollup should replace this when generating the package in `dist`

mxdvl (approved):
As we don’t actually run Rollup on `*.d.ts` files, I think that’s actually fine.

sndrs (approved):

mxdvl:
@ioanna0, if you have a single commit for a PR, the commit message is used instead of the PR title in `squash and merge`, annoyingly!

ioanna0:
@mxdvl Yes i know, only remembered when I merged and thus committed the empty fix message.

PR #399, opened 2021-01-19
fix: Switch debug console to new Error

ioanna0:
## What does this change?

Switch debug console to throwing a new Error when tcdata is empty

## Why?
Hopefully this would get picked up by Sentry


mxdvl (approved):
⚠️ LGTM 😉

PR #425, opened 2021-03-25
chore(tests): Run TCFv2 in CI workflow

mxdvl:
## What does this change?

Enables TCFv2 testing in continuous integration workflow.

## Why?

Sourcepoint now runs all frameworks globally, and it is the CMP’s job to make the right call on which framework to run. These tests were failing for a while, and it looks like #431 fixed them.


ioanna0 (approved):
👍 great!

PR #431, opened 2021-04-07
fix: Trigger callbacks when consent is ready

mxdvl:
## What does this change?

`onConsentReady` is when consent is really ready. We used `onMessageChoiceSelect`, but it’s not as reliable.

Shortly, we should simply add an event listener to the `__tcfapi`.

## Why?

New visits did not get a consented state and end-to-end tests are failing.


ashishpuliyel (approved):
Sweet. +1 For the record, we don't think `onMessageChoiceSelect` does anything useful at this point. However calling `invokeCallbacks` multiple times is safe because that function checks if the last value passed to the callbacks is different than the value that would be passed now.

PR #436, opened 2021-04-13
test: add end-to-end tests for aus

MarSavar:
Co-Authored-By: Max Duval <max.duval@theguardian.com>

## What does this change?
This adds end-to-end tests for Australia.

## Why?
They didn't exist before. We now have global geolocation enabled.


mxdvl:
When testing this locally, we couldn’t see the banner appear. The next step is to confirm that the Sourcepoint scenarios are set up to run globally for Australia.

mxdvl (approved):
We paired on this. It’s working well locally, and gives us more confidence in the future. Well done!

PR #442, opened 2021-05-06
fix: remove getting country code

ioanna0:
## What does this change?
This error wasn't pick up before

## Why?
gu.geolocation will become obsolete


mxdvl (approved):
Good shout.

PR #449, opened 2021-05-10
chore(deps): Bump dependencies 2021.05.10

mxdvl:
## What does this change?

Bump dependencies, including some major version bumps for:

- `@guardian/eslint-config`: 0.4 &rarr; **0.5**
- `@guardian/eslint-config-typescript`: 0.4 &rarr; **0.5**
- `@guardian/prettier`: 0.4 &rarr; **0.5**
- `@rollup/plugin-commonjs`: 16 &rarr; **19**
- `@rollup/plugin-node-resolve"`: 10 &rarr; **13**
- `@rollup/plugin-typescript"`: 6 &rarr; **8**
- `cypress"`: 5 &rarr; **7**
- `eslint-plugin-svelte3"`: 2 &rarr; **3**
- `prettier-plugin-svelte`: 1 &rarr; **2**

Use caret notation with only the major number, and let the robots handle the rest.

## Why?

We can't stay in the past.


mxdvl:
Updating Cypress causes issues with the end-to-end tests, where there seemingly is no framework on the page. I don't think that's quite right, however. Maybe a race condition?

ashishpuliyel (approved):
"The future is now" Love it

PR #472, opened 2021-06-21
docs: AUS Readme

mxdvl:
## What does this change?

Clarifies language.

## Why?

It’s better to be precise and accurate.


mxdvl:
This is already in place…

PR #477, opened 2021-06-30
fix: Rollback stub in TCFv2

mxdvl:
## What does this change?

- reset stub to pre-unified one
- make @guardian/libs a peer dependency again

## Why?

v6.12.0 introduced a regression which meant that certain ads stopped working.


src/tcfv2/stub.js
```diff
@@ -1,297 +1,91 @@
...
-	]);
+							n.parameter,
+						);
+				},
+				!1,
+			));
+	};
+	// call `e` directly. See https://git.io/JJxcM
+	// 'undefined' != typeof module ? (module.exports = e) : e();
+	e();
```
zekehuntergreen, line 90: @mxdvl calling `e` here puts `__tcfapi` on the `window` so no need to set a timeout in init!

zekehuntergreen:
We've decided to forgo this change in favour of the upgrade from the legacy to unified implementation of Sourcepoint.

PR #478, opened 2021-07-01
fix: Update TCF web banner color

zekehuntergreen:
## What does this change?
Changes TCF banner color from `#052862` to `#052962`.
The same change for banners on other platforms and in other regions is here #392 

## Why?
To reflect the style currently configured in the Sourcepoint UI.


SiAdcock (approved):
💎 💎 💎 💎 💎 💎 💎 💎 💎

PR #485, opened 2021-07-20
feat!: Unified SourcePoint Implementation

zekehuntergreen:
## What does this change?
- Upgrades SourcePoint from [legacy](https://documentation.sourcepoint.com/web-implementation/web-implementation/web-implementation-legacy) implementation  to [unified](https://documentation.sourcepoint.com/web-implementation/web-implementation/multi-campaign-web-implementation).

## Why?
- As described [here](https://documentation.sourcepoint.com/web-implementation/general/sdk-overview#unified-sdks), the unified implementations offers the following advantages:
   - Allows SourcePoint to develop new features, maintain, and update our SDKs more quickly.
   - Easier support for new and updated compliance regimes 
   - Simpler and faster client implementations
   - Improved messaging performance.
- Simplify the CMP by reducing duplicate or similar code.

PR #498, opened 2021-08-13
Remove deprecated parameter isInUsa

zekehuntergreen:


mxdvl:
Quick note to say this is also a breaking change, like #485, so make sure to identify it as such.

We’re not precious about version numbers ;)

zekehuntergreen:
Moving [here](https://github.com/guardian/consent-management-platform/pull/511) so that the change is independent of #485

PR #512, opened 2021-09-08
chore!: conventional commit preset

zekehuntergreen:
## What does this change?
Add dependency [conventional-changelog-conventionalcommits](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-conventionalcommits) and add preset conventionalcommits to `.releaserc`

## Why?
This allows the use inclusion of `!` in commit messages to trigger major releases.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "@types/jest": "^26",
     "@typescript-eslint/eslint-plugin": "^4",
     "@typescript-eslint/parser": "^4",
+    "conventional-changelog-conventionalcommits": "^4.6.0",
     "cypress": "^8",
     "cypress-wait-until": "^1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -91,6 +91,7 @@
     "wait-for-expect": "^3"
   },
   "dependencies": {
-    "@guardian/libs": "^1"
+    "@guardian/libs": "^1",
+    "conventional-changelog-conventionalcommits": "^4.6.0"
```
sndrs, line 95: i think this is a dev dep?

zekehuntergreen: Ah yes, good call

sndrs (approved):
lgtm!

PR #515, opened 2021-09-09
chore!: Version 8.0.0

zekehuntergreen:
## What does this change?
Create tag and release CMP version 8 (willl auto version based on PR title)

## Why?
Version 7.0.0 could not be released when #512 was merged because versions 7.0.0-7.0.4 already existed (see https://github.com/guardian/consent-management-platform/runs/3555099654). 
`npm view --json` shows that we skipped from 6.4.0 to 7.0.0 on 2020-10-20, then from 7.0.4 to 6.5.0 the next day.


arelra (approved):
!

PR #530, opened 2021-11-01
 fix: Fix CCPA for Unified Implementation 

zekehuntergreen:
## What does this change?
Fixes a regression introduced in version 9.0.0 which breaks advertising in CCPA regions by adding both __tcfapi and __uspapi functions to the window object. Advertisers assume that if __tcfapi is defined, it should take precedence over __uspapi.
This pull request fixes the issue by:
- Running either TCFv2 or CCPA stub file depending on framework so that before `wrapperMessagingWithoutDetection.js` is run, only one of the functions is defined.
- Adding either gdpr or ccpa object to _sp_ depending on framework. `wrapperMessagingWithoutDetection.js` uses the presence of these keys to determine which function to attach to `window`.
Note that neither of these changes is recommended by SourcePoint in their documentation. They are necessary because this consent management platform determines framework based on the country code passed to `init` rather than relying on SourcePoint's geolocation.

### DIFF
--- a/src/sourcepoint.ts
+++ b/src/sourcepoint.ts
@@ -26,7 +26,7 @@ const getProperty = (framework: Framework): Property => {
 };
 
 export const init = (framework: Framework, pubData = {}): void => {
-	stub();
+	stub(framework);
 
 	// make sure nothing else on the page has accidentally
 	// used the `_sp_` name as well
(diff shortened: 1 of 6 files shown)


mxdvl (approved):
Nice!

I think if we’re doing this it’s worth splitting the stubs in their individual files, and keep the caller in TypeScript land:

- `stub.ts`
- `stub_ccpa.js`
- `stub_tcfv2.js`

src/sourcepoint.ts
```diff
@@ -155,13 +154,21 @@ export const init = (framework: Framework, pubData = {}): void => {
 		},
 	};
 
+	// NOTE - Contrary to the SourcePoint documentation, it's important that we add EITHER gdpr OR ccpa
```
kenoir, line 157: Can we link to the relevant bit of the (incorrect) Sourcepoint documentation here? It might be handy for future folk working on this code to understand the context.

mxdvl (approved):

PR #535, opened 2021-11-23
fix: Wait for user action to trigger callbacks in TCFv2

mxdvl:
## What does this change?

Wait for user action to trigger callbacks in TCFv2. If `eventStatus` is `'cmpuishown'`, don’t trigger the callbacks.

## Why?

Otherwise, we always get a state similar to “Reject All” as the first state, very shortly followed by the actual one chosen by the user.

### DIFF
--- a/src/onConsentChange.test.js
+++ b/src/onConsentChange.test.js
@@ -186,4 +186,33 @@ describe('under TCFv2', () => {
 			expect(callback).toHaveBeenCalledTimes(2);
 		});
 	});
+
+	it('invokes callbacks only if there is a user action', async () => {
+		const callback = jest.fn();
+
+		tcData.eventStatus = 'cmpuishown';
+
+		onConsentChange(callback);
+		invokeCallbacks();
+
+		await waitForExpect(() => {
+			expect(callback).toHaveBeenCalledTimes(0);
+		});
+
+		tcData.eventStatus = 'useractioncomplete';
+
+		invokeCallbacks();
+
+		await waitForExpect(() => {
+			expect(callback).toHaveBeenCalledTimes(1);
+		});
+
+		tcData.eventStatus = 'tcloaded';
+
+		invokeCallbacks();
+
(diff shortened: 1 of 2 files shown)


arelra:
I agree with the change in principle but I'm just wondering if consumers of `onConsentChange` should be made aware?

Previously consumers could have relied on the initial state (probably unknowingly) for the first render.

However with this change it is now possible that the user will not interact for an unknown amount of time - or perhaps never.

In which case consumers of `onConsentChange` should block their dependant rendering until consent is provided.

They should do this anyway but with this change it _might_ mean any issues will be made apparent.

mxdvl:
@arelra should we make this a breaking change, then?

It’s worth noting that most people will still be on v6, and this is a fix for v9. As we’re not yet using this version anywhere, I would be keen to fix it before it rolls out.

arelra:
> @arelra should we make this a breaking change, then?
> 
> It’s worth noting that most people will still be on v6, and this is a fix for v9. As we’re not yet using this version anywhere, I would be keen to fix it before it rolls out.

Ah okay. It's a  _potential_ breaking change for consumers.

If no one is using v9 it could be part of that release.

Would be good to give advance warning to CMP clients though.

shtukas:
I am ok with approving the code change if needed, just wondering if we reached a conclusion after the above discussion...

ashishpuliyel:
I was speaking to @mxdvl about this and I can provide a bit of context. When this was originally implemented at the time there was a fear that all consumers wouldn't necessarily handle the null/undefined/pending-consent state, so this was done to default to the no-consent state, which we knew would be a 'safe' handled state.
It was simply defensive programming.

Now we have more confidence about the behaviour of consumers of the CMP, we know that they understand they need to wait for consent, and until then not fire things that require consent.

mxdvl:
So I think we’re happy with this going in `v9`. Just need a ✅  🙏

arelra (approved):

PR #540, opened 2021-12-31
feat: Define and export type `VendorName`

zekehuntergreen:
## What does this change?
- Export type `VendorName`, an enum of the vendors supported by the CMP (`'a9' | 'acast' | 'braze' `, etc.)

## Why?
`VendorName` needs to be imported by consumers of `getConsentFor` so that they can be converted to TypeScript.


ashishpuliyel (approved):
Perfect!

PR #544, opened 2022-01-24
fix: Prebid uses IAB vendor ID

mxdvl:
## What does this change?

We’ve now switched to the official Prebid vendor ID.

## Why?

We can’t check for Prebid consent otherwise.


ashishpuliyel (approved):
Perfect

shtukas:
+1

PR #548, opened 2022-02-09
[DO NOT MERGE] How should I read this enum?

kenoir:
This is a draft PR to help pose a question around accessing an `enum` inside `./src` from a `.ts` script outside that directory. It's intended to serve as an example and should not be merged.


kenoir:
Summary: Don't do this, it is overly complicated - do it in `getConsentFor.test.ts` instead. Thanks @mxdvl.

PR #549, opened 2022-02-10
ci: Add test to check Vendor IDs are in the Sourcepoint Guardian vendor list

kenoir:
## What does this change?

This adds a test that should run in the CI environment that compares the list of vendor IDs we hard-code in `getConsentFor.ts` to those published on the IAB `consensu.org` domain that contains the Guardian vendor list.

## Why?

If we update out vendor list with our CMP supplier, but do not update the list in this repository when vendors change IDs they can become out of date here. Subsequently when we check consent for a particular vendor using `getConsentFor` it will not be available and that vendor will fail to run.

This test is intended to catch any mismatch between the list encoded here and that registered with our CMP as soon as possible.

[Trello](https://trello.com/c/UOC47dKN/598-add-a-ci-check-to-the-cmp-that-checks-the-public-list-of-vendors-for-our-web-vendor-list-vs-the-list-in-the-cmp)


mxdvl (commented):
Automated tests!

You might already have `node-fetch` instead of adding `axios`.

kenoir:
This PR is currently failing because the CI test is picking out some vendors that are missing.  These are:

- `'facebook-mobile' = '5e716fc09a0b5040d575080f'`
- `'firebase' = '5e68dbc769e7a93e0b25902f'`
- `'google-sign-in' = '5e4a5fbf26de4a77922b38a6'`

Searching [our org](https://github.com/search?p=4&q=org%3Aguardian+getConsentFor&type=Code) I cannot see where these `getConsentFor` is called for these vendors so I presume it is safe to presume them?

kenoir:
> Automated tests!
> 
> You might already have `node-fetch` instead of adding `axios`.

We don't seem to have `node-fetch` so i'm going to keep `axios`.

PR #552, opened 2022-02-15
fix: Remove vendors ids not registered with Sourcepoint

kenoir:
## What does this change?

Following https://github.com/guardian/consent-management-platform/pull/549 we seem have some vendor IDs we are not using (and are not currently registered in our vendor list configuration with Sourcepoint).

## Why?

These should be removed to keep the` VendorID` enum up to date and allow the CI test mentioned above to run without failure.


ashishpuliyel (approved):
Repeating what I said to you on chat here, for the record:

This PR makes sense. I assume no team is calling getConsentFor on them.
It shouldn't make a difference. We should just make it so calling getConsentFor for an ID that doesn't exist on SP is the same as calling getConsentFor for a name that doesn't exist in 'our' list. Both return undefined and log an error.

PR #555, opened 2022-02-18
feat: Make initialising the CMP server-safe

kenoir:
## What does this change?

This PR attempts to make initialising the CMP in a server environment (where there is no `window`) safe to do.

## Why?

From the suggestion in to https://github.com/guardian/dotcom-rendering/pull/3925

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -16,9 +16,13 @@ import type {
 } from './types';
 import type { Country } from './types/countries';
 
+const isServerSide = typeof window === 'undefined';
+
 // Store some bits in the global scope for reuse, in case there's more
 // than one instance of the CMP on the page in different scopes.
-window.guCmpHotFix ||= {};
+if (!isServerSide) {
+	window.guCmpHotFix ||= {};
+}
 
 let frameworkCMP: SourcepointImplementation | undefined;
 
@@ -26,12 +30,13 @@ let _willShowPrivacyMessage: undefined | boolean;
 let initComplete = false;
 
 let resolveInitialised: (value?: unknown) => void;
+
 const initialised = new Promise((resolve) => {
 	resolveInitialised = resolve;
 });
 
 const init: InitCMP = ({ pubData, country }) => {
-	if (isDisabled()) return;
+	if (isDisabled() || isServerSide) return;
 
 	if (window.guCmpHotFix.initialised) {
 		if (window.guCmpHotFix.cmp?.version !== __PACKAGE_VERSION__)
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -104,21 +111,50 @@ const showPrivacyManager = () => {
...
-
-	// special helper methods for disabling CMP
-	__isDisabled: isDisabled,
-	__enable: enable,
-	__disable: disable,
-});
+export const cmp: CMP = (() => {
+	if (isServerSide) {
+		return {
+			init: (_: { pubData?: PubData; country?: Country }) => void 0,
```
sndrs, line 117: do you need this?

```suggestion
			init: () => void 0,
```

src/index.ts
```diff
@@ -104,21 +111,50 @@ const showPrivacyManager = () => {
...
+			showPrivacyManager: () => {
+				void 0;
+			},
```
sndrs, line 120: just for consistency
```suggestion
			showPrivacyManager: () => void 0,
```

src/index.ts
```diff
@@ -104,21 +111,50 @@ const showPrivacyManager = () => {
...
+export const onConsentChange = (() => {
+	if (isServerSide) {
+		return () => void 0;
+	}
+	return (window.guCmpHotFix.onConsentChange ||= actualOnConsentChange);
+})();
+
+export const getConsentFor = (() => {
+	if (isServerSide) {
+		actualGetConsentFor;
```
sndrs, line 157: should this be another one of these?

```suggestion
		return () => void 0;
```

kenoir: As far as I can tell the implementation of `getConsentFor` is stateless and already safe if it were without `window`.

src/index.ts
```diff
@@ -14,22 +14,27 @@ import type {
 	WillShowPrivacyMessage,
 } from './types';
 
+const isServerSide = typeof window === 'undefined';
+
 // Store some bits in the global scope for reuse, in case there's more
 // than one instance of the CMP on the page in different scopes.
-window.guCmpHotFix ||= {};
+if (!isServerSide) {
+	window.guCmpHotFix ||= {};
```
DavidLawes, line 22: Sorry, probably a complete aside, but I found it really difficult to reason around the use of `||=` - maybe just my lack of familiarity though!

oliverlloyd: Oh yes, me too. I read the docs for this but _still_ haven't got a mental model for it

kenoir: I'm thinking about it like this:

```
window.guCmpHotFix = window.guCmpHotFix || {}
```

sndrs: this used to be a very common pattern, similar to @kenoir's example:

`window.guCmpHotFix || (window.guCmpHotFix = {})`

it replaces that iiuc

OllysCoding (approved):
+1 Looks good to me! Tested locally on DCR & with cmp included in the server side bundle, everything worked, no more 'window is undefined' errors :)

mxdvl:
Did something get in the way of automatic releases?

PR #559, opened 2022-03-01
fix: Simplify server side implementation

kenoir:
## What does this change?

Incorporate changes from https://github.com/guardian/dotcom-rendering/pull/4106

## Why?

This change makes it clear the CMP is not to be used server-side and warns in console to that effect.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,10 +2,16 @@ import { log } from '@guardian/libs';
 import { AUS } from './aus';
 import { CCPA } from './ccpa';
 import { disable, enable, isDisabled } from './disable';
-import { getConsentFor as actualGetConsentFor } from './getConsentFor';
+import { getConsentFor as clientGetConsentFor } from './getConsentFor';
 import { setCurrentFramework } from './getCurrentFramework';
 import { getFramework } from './getFramework';
-import { onConsentChange as actualOnConsentChange } from './onConsentChange';
+import { onConsentChange as clientOnConsentChange } from './onConsentChange';
+import {
+	isServerSide,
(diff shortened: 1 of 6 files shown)


src/index.ts
```diff
@@ -112,17 +124,17 @@ const showPrivacyManager = () => {
...
+			__isDisabled: serverSideWarnAndReturn(false),
+			hasInitialised: serverSideWarnAndReturn(false),
+			init: serverSideWarn,
+			showPrivacyManager: serverSideWarn,
+			version: 'n/a',
+			willShowPrivacyMessage: serverSideWarnAndReturn(
+				Promise.resolve(false),
+			),
+			willShowPrivacyMessageSync: serverSideWarnAndReturn(false),
 		} as CMP;
```
mxdvl, line 138: You should be able to remove the assertion!
```suggestion
		};
```

mxdvl (commented):
I’d be tempted to move these methods to their own file?

It seems `index.ts` is starting to get quite busy.
If you adopt a pattern where you’re putting the values in a `server.ts` file for example, it might allow consumers to have the same kind of logic we have in dotcom-rendering, where the alias is actually using the server-side imports directly.


```ts
import { onConsentChange as clientOnConsentChange } from './onConsentChange';
import { onConsentChange as serverOnConsentChange } from './server';

export const onConsentChange = isServerSide ? serverOnConsentChange : window.guCmpHotFix.onConsentChange ||= clientOnConsentChange;
```

shtukas (approved):

mxdvl (approved):
This is looking great!

PR #560, opened 2022-03-01
feat: Wait for user action to trigger callbacks in TCFv2

arelra:
## What does this change?

Re-applying a previous fix from #535 

Wait for user action to trigger callbacks in TCFv2. If `eventStatus` is `'cmpuishown'`, don’t trigger the callbacks.

## Why?

Otherwise, we always get a state similar to “Reject All” as the first state, very shortly followed by the actual one chosen by the user.


src/onConsentChange.test.js
```diff
@@ -185,4 +185,33 @@ describe('under TCFv2', () => {
 			expect(callback).toHaveBeenCalledTimes(2);
 		});
 	});
+
+	it('invokes callbacks only if there is a user action', async () => {
```
kenoir, line 189: My understanding is that `onConsentChange` will still be called in the case we have stored a previous valid consent choice, so not _only_ if there is a user action.

Should this be something like `'if cmpuishown callbacks are invoked only after user action'`?

PR #562, opened 2022-03-02
feat: Allow multiple sourcepoint IDs for each vendor ID

kenoir:
## What does this change?

Allow the library to specify multiple Sourcepoint IDs to be looked up for a vendor ID, and then check if those IDs are in the consent state.

## Why?

This is intended to allow "new" and "old" ids to be present in the CMP library when vendors move from custom to IAB status (or any other ID changing moving). So that the CMP can be deployed for a vendor list update and transition seamlessly to using the new ID.

For https://trello.com/c/loSA82Sy/597-add-a-mechanism-to-the-cmp-to-transition-from-old-to-new-ids-seamlessly-without-interruption-of-service-during-a-re-permission


kenoir:
Replaces https://github.com/guardian/consent-management-platform/pull/556

shtukas (approved):

ashishpuliyel (approved):
This is neat Kenny. Thanks for doing this.

zekehuntergreen (approved):
Looks good!

PR #577, opened 2022-03-14
[DRAFT] Adds monitoring lambda checks to CMP

kenoir:
## What does this change?

This PR adds infrastructure and code to run the same checks as the [Commercial Canaries](https://github.com/guardian/aws-canaries/tree/main/src) against an arbitrary environment (e.g. local, CODE, PROD), so these checks can be run as part of the deployment process before reaching the production environment.

These checks will run in lambdas in AWS or can be run locally against a local instance of DCR. 

This change borrows the setup from the [ariadne](https://github.com/guardian/ariadne) project to package a lambda that runs puppeteer 

## Why?

Currently when deploying the CMP we will run very similar tests to the [Commercial Canaries](https://github.com/guardian/aws-canarieshttps://github.com/guardian/aws-canaries) manually and against a branch of DCR & Frontend deployed in the CODE environment.

Instead of waiting for errors in the production environment or manually performing a test script locally/in CODE that can be forgotten or done incorrectly we should encode our manual tests and run them on-demand in AWS in each region we wish to test the CMP. 

[AWS Synthetics](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Synthetics_Canaries.html) is not appropriate for this purpose as it can only be run on a schedule.


monitoring/src/check-page/tcfv2.ts
```diff
@@ -94,3 +99,287 @@ export const checkPage = async function (
...
+
+const loadPage = async (page: Page, url: string) => {
+	log_info(`Loading page: Start`);
+
+	const response = await page.goto(url, {
+		waitUntil: 'domcontentloaded',
+		timeout: 30000,
+	});
+
+	// For some reason VSCode thinks the conditional is not needed, because `!response` is always falsy 🤔
```
kenoir, line 285: I came across this same issue when adapting this code the first time (see `checkPage_old`). I think you can safely remove this statement. `page.goto` will throw if there is an issue rather than return something falsy.

monitoring/src/check-page/aus.ts
```diff
@@ -0,0 +1,204 @@
...
+	checkCMPIsOnPage,
+	checkTopAdHasLoaded,
+	clearCookies,
+	loadPage,
+	log_error,
+	log_info,
+	makeNewBrowser,
+} from './common-functions';
+
+// -- Original code from Robert ------------------------------------------------
```
kenoir, line 13: I think we should remove the commented out code here referencing the other implementations.

monitoring/src/check-page/aus.ts
```diff
@@ -0,0 +1,204 @@
...
+
+	if (nextUrl) {
+		await checkSubsequentPage(nextUrl);
+	}
+};
+
+export const mainCheck = async function (): Promise<void> {
+	log_info('checkPage, new version (aus)');
+	await checkPages(
+		'https://www.theguardian.com/au?adtest=fixed-puppies',
```
kenoir, line 197: We are not paying attention to the configured environment here.

monitoring/src/check-page/aus.ts
```diff
@@ -0,0 +1,204 @@
...
+const interactWithCMP = async (page: Page) => {
+	// Ensure that Sourcepoint has enough time to load the CMP
+	await page.waitForTimeout(5000);
+
+	// When AWS Synthetics use a more up-to-date version of Puppeteer,
+	// we can make use of waitForFrame(), and remove the timeout above.
+	log_info(`Clicking on "Continue" on CMP`);
+	const frame = page
+		.frames()
+		.find((f) => f.url().startsWith('https://ccpa-notice.sp-prod.net'));
```
kenoir, line 119: This value should be loaded out of the config (see how the commented out code does this).

monitoring/src/check-page/aus.ts
kenoir: I think this problem persists in all 3 region files.

monitoring/src/check-page/common-functions.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export const loadPage = async (page: Page, url: string): Promise<void> => {
+	log_info(`Loading page: Start`);
+
+	const response = await page.goto(url, {
+		waitUntil: 'domcontentloaded',
+		timeout: 30000,
+	});
+
+	// For some reason VSCode thinks the conditional is not needed, because `!response` is always falsy 🤔
```
kenoir, line 86: I think there was a resolution to this in the first iteration of this code, might be worth looking at how this function was implemented there.

monitoring/src/config.ts
```diff
@@ -0,0 +1,115 @@
...
+	jurisdiction: 'aus',
+	frontUrl: 'https://www.theguardian.com/au',
+	articleUrl:
+		'https://www.theguardian.com/food/2020/dec/16/how-to-make-the-perfect-vegetarian-sausage-rolls-recipe-felicity-cloake',
+	iframeDomain: 'https://ccpa-notice.sp-prod.net',
+	checkFunction: mainCheckAus,
+};
+
+const ConfigAusCode: Config = {
+	stage: 'code',
```
kenoir, line 82: The `STAGE` environment variables (at least in CODE) that are handed to the lambda are uppercase so the stage matching below does not work and the tests are run against the production environment instead.

Perhaps we need to think about how this parameter will be set when we call this lambda?

monitoring/src/index.ts
```diff
@@ -0,0 +1,21 @@
...
+	);
+
+	try {
+		const browser: Browser = await run(envConfig, debugMode);
+
+		if (!debugMode) {
+			await browser.close();
+		}
+		console.log('(cmp monitoring) Finished cmp-monitoring');
+	} catch (error) {
```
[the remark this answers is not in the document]
shtukas: Yeah. It was mostly stylistic for when I was running on local

monitoring/src/config.ts
kenoir: I think this might be the case because I can see prod config urls being called this way in the eu-west-1 CODE deployment. As discussed with @shtukas, possibly there is another cause.

shtukas:
Hi Robert, I corrected the timeout you observed last week (was a stupid mistake from my part). 

In the code used by commercial dev, the signature of `checkSubsequentPage` is 

```
const checkSubsequentPage = async (url)
```

but in our case we also needed to pass the browser

```
const checkSubsequentPage = async (
    browser: Browser,
    config: Config,
    url: string,
)
```

In the previous version of the code, before my last commit, `checkSubsequentPage` was rebuilding the browser and starting with a fresh set of cookies, thereby not inheriting the works done in `checkPages`. In the last commit, I simply pass the browser itelf to `checkSubsequentPage`, and that corrected the problem.

PR #593, opened 2022-04-05
feat: Adds linkedin to list of vendors

kenoir:
## What does this change?

This change adds `linkedin` to the list of vendors so that a call to `getConsentFor` can reference this vendor when checking for consent.

The Sourcepoint vendor ID mapping is [retrieved from their API](https://sourcepoint.mgr.consensu.org/tcfv2/vendor-list?vendorListId=5ec67f5bb8e05c4a1160fda1).

## Why?

This is required for to allow the marketing team responsible for Membership events to add LinkedIn's Insight tracking pixel and have this be added to the page _only_ when consent has been given.


GHaberis (approved):
Thanks 👍

Georges-GNM (approved):

PR #606, opened 2022-05-20
feat: Notice & Report on JS GPC signal

kenoir:
## What does this change?

This change adds `gpcSignal` to `ConsentState` indicating the presence of the [GPC JavaScript property to detect preference](https://globalprivacycontrol.github.io/gpc-spec/#javascript-property-to-detect-preference).

It does **not** change the state of `canTarget` (yet).

## Why?

This change will allow us to report on and better respond to the GPC signal sent by browser.


shtukas (approved):
LGTM!

mxdvl (commented):
Do we want `true`, `false` and `undefined`, or should we convert this to a boolean?

In JS, if a key is set to the same name as an in-scope variable, you can use the shorthand:

```ts
const gcpSignal = getGcpSignal();
const state = {
  gcpSignal, // same as `gcpSignal: gcpSignal,`
}
```

kenoir:
> Do we want `true`, `false` and `undefined`, or should we convert this to a boolean?

`undefined` is interesting as it's useful to know where we are seeing browsers providing the variable on `navigator` at all. For example in Chrome this will be undefined (without a special extension), but in Firefox with GPC enabled this will be defined.

kenoir:
@mxdvl the intention is to update [commercial-core](https://github.com/guardian/commercial-core/blob/main/src/send-commercial-metrics.ts) to read this metric and report on it.

PR #615, opened 2022-06-08
test: Add simple script to invoke lambda in available regions

kenoir:
## What does this change?

Adds a script to remotely invoke the monitoring lambdas in the available global regions.

## Why?

So we can see what their results are!


shtukas (approved):
🎉

PR #620, opened 2022-06-16
test: Fix problem with multiple Browser instances

kenoir:
## What does this change?

This fixes runtime problems in AWS. The problem was caused by incorrectly spawning multiple (up to 3) Browsers without correctly closing them after use. In particular

1. The browser originally created in index.ts, was useless.
2. The browser created in the checkPages function was not being closed (resulting in two instances running)

The browser in (1.) and the two browsers in (2.) contributed to three instances running at the same time. 

And while being at it, we also simplified the code.

Co-authored-by: @akinsola-guardian

### DIFF
--- a/monitoring/src/index.ts
+++ b/monitoring/src/index.ts
@@ -1,5 +1,4 @@
 import { envConfig } from './config';
-import { run } from './puppeteer';
 
 export const handler = async (): Promise<void> => {
 	console.log(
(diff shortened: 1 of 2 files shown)


shtukas:
👍

shtukas:
Addition in https://github.com/guardian/consent-management-platform/pull/624

PR #625, opened 2022-06-22
test: Re-enable debug-mode & fix clearing local storage

kenoir:
## What does this change?

The most recent version of the Sourcepoint code uses LocalStorage to store state - we need to update our tests to reset this when appropriate.

In addition, this change reinstates `DEBUG_MODE` that became broken in a previous changeset.

## Why?

We want to be able to have reliable tests!


shtukas (approved):

kenoir:
Related commercial canaries PR: https://github.com/guardian/aws-canaries/pull/20

PR #630, opened 2022-06-23
feat: Automate CMP Testing

akinsola-guardian:
## What does this change?
- Automates CMP tests using AWS Eventbridge for scheduled triggers.
- Refactored allow user to trigger lambda in different environments and jurisidictions.

## Why?
When we deploy the CMP to CODE we want to test if it is behaving as expected, we currently do this manually. However there are automated tests against PROD in the form of the "commercial canaries".

We should adapt the current tests to run against the CODE environment on demand to improve speed and reliability of CMP deployment.


monitoring/src/config.ts
```diff
@@ -106,19 +84,53 @@ const availableEnvConfig = [
...
+	async run(): Promise<void> {
+		this.generateConfig();
+		if (this._config) {
+			await this._config.checkFunction(this._config);
+		}
 	}
 
-	return foundConfig;
-})();
+	private generateConfig(): void {
```
kenoir, line 111: Very nice. Maybe make it log what it's doing i.e. `console.log("Running in Scheduled mode")` or something like that.

monitoring/src/utils/config-helper.ts
```diff
@@ -0,0 +1,56 @@
...
+	AP_SOUTHEAST_2: 'ap-southeast-2',
+};
+
+export const JURISDICTIONS = {
+	TCFV2: 'tcfv2',
+	CCPA: 'ccpa',
+	AUS: 'aus',
+};
+
+export class ConfigHelper {
```
kenoir, line 16: Does this need to be a class or can we export the functions without the class wrapper?

monitoring/src/utils/logger.ts
```diff
@@ -0,0 +1,75 @@
+class Logger {
```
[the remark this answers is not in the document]
kenoir: There is also this in [`guardian/libs`](https://github.com/guardian/libs/blob/main/src/logger/README.md) but that is intended as a client side logger (linked here for reference as it's still interesting).

akinsola-guardian:
[Run the lambdas on a schedule](https://trello.com/c/wcRBLzRl/708-run-the-lambdas-on-a-schedule)

PR #635, opened 2022-07-01
feat: Automate-CMP-Testing

akinsola-guardian:
### **What does this change?**

- Automates CMP tests using AWS Eventbridge for scheduled triggers.
- Refactored allow user to trigger lambda in different environments and jurisdictions.
- Adds tests to the monitoring directory

###  Why?
When we deploy the CMP to CODE we want to test if it is behaving as expected, we currently do this manually. However there are automated tests against PROD in the form of the "commercial canaries".

We should adapt the current tests to run against the CODE environment on demand to improve speed and reliability of CMP deployment.


akinsola-guardian:
[Run the lambdas on a schedule](https://trello.com/c/wcRBLzRl/708-run-the-lambdas-on-a-schedule)

monitoring/README.md
```diff
@@ -0,0 +1,11 @@
+# Automated Monitoring
```
kenoir, line 1: 👏

cdk/lib/monitoring.ts
```diff
@@ -15,14 +17,32 @@ export class Monitoring extends GuStack {
...
+		);
+
+		const lambdaEventTarget = new LambdaFunction(monitoringLambdaFunction, {
+			event: RuleTargetInput.fromObject({
+				stage: 'PROD', // Both scheduled cmp-monitoring-CODE and cmp-monitoring-PROD are monitoring prod versions
+				region: region,
+			}),
+		});
+
+		const monitoringDuration: number = stage === 'PROD' ? 2 : 5; // Every 5 minutes for CODE; Every 2 minutes for PROD.
```
kenoir, line 41: I think we probably don't care if this runs at all in CODE really, this could be once a day.

akinsola-guardian: Thanks, I've updated this,

monitoring/remoteRun.ts
```diff
@@ -32,18 +44,29 @@ async function main() {
...
-			processResult(result.value);
-		} else {
-			console.log('Failed to get response: ', result.reason);
-		}
+	await prompt([
+		{
+			type: 'list',
+			name: 'stage',
+			message: 'Which environment would you like to test?',
+			choices: ['prod', 'code'],
```
kenoir, line 52: If this is running locally, we may also want to run it against a version of frontend/dcr on my local machine. That is probably something to do in a future PR.

monitoring/localRun.ts
```diff
@@ -1,6 +1,44 @@
...
-	handler().catch(console.error);
+
+	await prompt([
+		{
+			type: 'list',
+			name: 'stage',
+			message: 'Which environment would you like to test?',
+			choices: ['code', 'prod'],
+		},
+		{
```
kenoir, line 19: I think if we run locally we never want to run against multiple regions, so in this case it should probably be a `list`.

monitoring/localRun.ts
```diff
@@ -1,6 +1,44 @@
...
+				return true;
+			},
+		},
+	]).then(async (userInput: LocalRunCLIUserInput) => {
+		for (const jurisdiction of userInput.jurisdictions) {
+			const event: CustomScheduleEventContent = {
+				stage: userInput.stage,
+				jurisdiction: jurisdiction,
+			};
+			await handler(event);
```
kenoir, line 38: As we're treating `localRun` like a CLI tool, if the test fails it would make sense to fail with a non-zero exit code if the test run fails.

monitoring/localRun.ts
```diff
@@ -1,6 +1,44 @@
...
 async function main() {
 	const { handler } = await import('./src');
-	handler().catch(console.error);
+
+	await prompt([
+		{
+			type: 'list',
+			name: 'stage',
+			message: 'Which environment would you like to test?',
+			choices: ['code', 'prod'],
```
kenoir, line 17: If this is running locally we may also want to add the option to run this against the URLs DCR/frontend presents when run locally.

monitoring/remoteRun.ts
```diff
@@ -32,18 +44,29 @@ async function main() {
...
+		},
+	]).then(async (userInput: RemoteRunCLIUserInput) => {
```
kenoir, line 54: I think we should be able to select the regions to check against locally here (defaulting to all of them).

monitoring/remoteRun.ts
```diff
@@ -32,18 +44,29 @@ async function main() {
...
+		const invokeSettledResults = await Promise.allSettled(
+			regionsToCheck.map((region) =>
+				invokeInRegion(region, 'cmp-monitoring-CODE', userInput.stage),
+			),
+		);
+
+		invokeSettledResults.map((result) => {
```
kenoir, line 61: Could we aggregate the overall results and display that result in a nice way on the command line (also same comment about exit status here).

monitoring/src/config.ts
```diff
@@ -106,19 +85,73 @@ const availableEnvConfig = [
...
+	}
+
+	public generateConfig(): void {
+		// If no jurisdiction assign using aws region (Scheduled)
+		if (!this._jurisdiction && this._awsRegion) {
+			this._jurisdiction = ConfigHelper.getJurisdiction(this._awsRegion);
+			log_info(`Generating config for scheduled trigger`);
+		}
+
+		// If no aws Region assign using jurisdiction (Adhoc)
```
kenoir, line 137: This makes me think we should probably be offering the region choice up front in the UI _not_ jurisdiction. Then the later issue of assigning region from jurisdiction goes away.

kenoir (commented):
Nice and slick - I like inquirer.

Some comments on functionality, but code and tests 👌

monitoring/localRun.ts
```diff
@@ -1,6 +1,44 @@
...
+type LocalRunCLIUserInput = {
+	stage: string;
+	jurisdictions: string[];
+};
+
 async function main() {
 	const { handler } = await import('./src');
-	handler().catch(console.error);
+
+	await prompt([
```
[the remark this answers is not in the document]
akinsola-guardian: I used commander to take arguments via cli. it falls back to the interactive cli if the arguments are invalid or 'yarn start' is called.

kenoir (approved):

PR #638, opened 2022-07-05
feat: Set up alerting on CMP monitoring failure.

akinsola-guardian:
## What does this change?

-  Adds metrics for invocations and errors
- Throws an error instead of returning  a failure key.
## Why?
To allow us to monitor the failures of the test

https://trello.com/c/lapOlaOY


shtukas (approved):
LGTM

PR #648, opened 2022-07-21
refactor: generate stacks from list of envs and regions

kenoir:
## What does this change?

Generate new stacks based on lists of region & env.

## Why?

Removes a bit of code duplication, makes adding removing regions a bit more straightforward.

note: this will rename the Asia Pacific region stack (replacing AU with AP)

### DIFF
--- a/cdk/lib/monitoring.test.ts
+++ b/cdk/lib/monitoring.test.ts
@@ -1,6 +1,4 @@
-import '@aws-cdk/assert/jest';
-import { SynthUtils } from '@aws-cdk/assert';
-import { App } from '@aws-cdk/core';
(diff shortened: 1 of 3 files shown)


cdk/lib/monitoring.test.ts
```diff
@@ -1,6 +1,6 @@
 import '@aws-cdk/assert/jest';
 import { SynthUtils } from '@aws-cdk/assert';
-import { App, Stage } from '@aws-cdk/core';
```
kenoir, line 3: `Stage` was unused.

dominickendrick (approved):

PR #655, opened 2022-08-09
fix: bump version of guardian/lib

sookburt:
## What does this change?
Updates to the latest version of guardian/lib

## Why?
To keep up to date.

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,6 +95,6 @@
     "wait-for-expect": "^3"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


shtukas (approved):

kenoir:
@sookburt it would be a good idea to test this version of the CMP locally in DCR & Frontend as this is a significant version jump. You can do so using `npm link`.

dominickendrick (approved):
Tested locally with @sookburt

PR #656, opened 2022-08-11
chore: Flaky cypress test improvement

Georges-GNM:
## What does this change?
tries to make tests more reliable and bumps the version of cypress to the latest 10.4.0. 

Unfortunately still seems to fail 1 in a 100 runs (possibly due to an issue with the sourcepoint site that gets fetched when the test is run), but on discussion with Dominic Kendrick, this sounds like the best that we can do for now.

## Why?

Prompted by this card: https://trello.com/c/32Y2NPdi


kenoir (approved):
👍 ... and with bonus linting!

It might have been neater to split that work into another PR but it's great to add it.

PR #664, opened 2022-09-15
feat: Add `final` parameter to onConsentChange

zekehuntergreen:
## What does this change?
- adds an optional `final` parameter to `onConsentChange` which defaults to `false`. When a true value is passed, the callback is pushed onto a new `finalCallbackQueue`.  The `finalCallbackQueue`'s callbacks are guaranteed to execute after the default `callBackQueue`'s when consent state changes.

## Why?
- In https://github.com/guardian/frontend/pull/25409 we use `onConsentChange` to register a callback that reloads the page when consent state changes from a state where the user has accepted targeted advertising to one which rejects it or vice versa.
- reloading the page as part of an `onConsentChange` callback opens up the possibility of overriding other uses of `onConsentChange` whose callbacks won't have time to execute before the page is reloaded. This could cause bugs that are very difficult to track down.
- Another approach considered was to wrap https://github.com/guardian/frontend/pull/25409 's callback in `Promise.resolve().then` and / or `setTimeout` so that the callback that reloads the page is put on the microtask queue or task queue respectively. This would increase the probability but not guarantee that other callbacks registered with `onConsentChange` could execute before the page is reloaded.


Mark-McCracken:
According to @sookburt , @theemilyturner asked Susan Bingham about whether we should reload the page for a consent change, and the answer was no, so this is no longer relevant.
If no-one has anything to say to the contrary, and if still open in a week, we'll close this.

Mark-McCracken:
as evidenced here: https://trello.com/c/sqmNkN20

sookburt:
This might need to be resurrected as part of the cookie deny list changes.

akinsola-guardian (approved):
Thanks for this. This was tested in CODE and worked as expected.

PR #667, opened 2022-09-27
refactor: Moving from using title text to using apparently-unique class names to identify SourcePoint buttons

sookburt:
## What does this change? 

Alters the monitoring test to use class names to identify the buttons within Sourcepoint's CMP in the absence of HTML element ids.  

## Why?

This fixes the issue that ab tests often change the text of the button which we were previously relying on.  It makes the monitoring tests slightly less fragile until we can persuade Sourcepoint to add ids into their code. (we hope).

Sue Burt and Akinsola Lawanson

### DIFF
--- a/monitoring/src/check-page/aus.ts
+++ b/monitoring/src/check-page/aus.ts
@@ -11,7 +11,7 @@ import {
 	makeNewBrowser,
 } from './common-functions';
 
-const interactWithCMP = async (config: Config, page: Page) => {
+const clickAcceptAllCookies = async (config: Config, page: Page) => {
 	// Ensure that Sourcepoint has enough time to load the CMP
 	await page.waitForTimeout(5000);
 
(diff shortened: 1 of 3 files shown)


monitoring/src/check-page/aus.ts
```diff
@@ -23,7 +23,9 @@ const interactWithCMP = async (config: Config, page: Page) => {
 		return;
 	}
 
-	await frame.click('button[title="Continue"]');
+	await frame.click(
+		'div.message-component.message-row > button.sp_choice_type_11',
```
kenoir, line 27: This seems like it might be quite brittle if SP change their implementation and it makes it hard to understand what this line of code is doing.

Are we making these changes to allow for the AB tests changing the language? If so, this makes more sense, but I think it would still be good to move these `frame.click`s into functions with more descriptive names or add comments with detail about what the click is supposed to be doing

Georges-GNM: In essence yes - this change would bring these tests in line with the latest [commercial canaries update](https://github.com/guardian/commercial-canaries/pull/35) (on the back of the AB test) which now use the CSS handle, which as far as I can tell is something we define in the SP interface. 
I agree that adding a comment stating what the click is doing would be helpful though.

kenoir (approved):

monitoring/src/check-page/aus.ts
sookburt: I've altered the method names to make them clearer.

PR #659, opened 2022-08-20
refactor: Switch to using an auto-generated riff-raff.yaml file

akash1810:
Part of https://github.com/guardian/consent-management-platform/issues/645.

## What does this change?
In this change we move to an auto-generated `riff-raff.yaml` file using @guardian/cdk, a feature released in [v47.1.0](https://github.com/guardian/cdk/releases/tag/v47.1.0).

## Why?
Primarily, this is to test the experimental[^1] GuCDK feature "in the wild". But also, it simplifies things a little; should we wish to deploy to another region, we'd now just need to update the [`cdk.ts` file](https://github.com/guardian/consent-management-platform/blob/6c59ef6692b3635ea2302c034d4f62c8df5ff04c/cdk/bin/cdk.ts#L8-L13) and the generated `riff-raff.yaml` file will include the new region.

## Notes
The manual `riff-raff.yaml` explicitly names the CloudFormation stack to deploy to, whereas the auto-generated file opts to use tag discovery (Stack, Stage, and App tags).

~We need to ensure these tags are present on the CloudFormation stacks _before_ shipping this, else Riff-Raff will try to create a new CFN stack ([and fail](https://riffraff.gutools.co.uk/deployment/view/c453d665-c476-41a3-b674-a061e49ae1f2)), which is undesired!~ Update: Riff-Raff is applying the standard SSA tags to CloudFormation stacks since https://github.com/guardian/riff-raff/pull/934. That is, the aforementioned concern is no longer present.

<details>
<summary>Content of generated file</summary>

```yaml
allowedStages:
  - CODE
  - PROD
deployments:
  lambda-upload-eu-west-1-cmp-monitoring-cmp-monitoring-lambda-eu-west-1:
    type: aws-lambda
    stacks:
      - cmp-monitoring
    regions:
      - eu-west-1
    app: cmp-monitoring-lambda-eu-west-1
    contentDirectory: cmp-monitoring-lambda-eu-west-1
    parameters:
      bucketSsmLookup: true
      lookupByTags: true
      fileName: cmp-monitoring-lambda-eu-west-1.zip
    actions:
      - uploadLambda
  cfn-eu-west-1-cmp-monitoring-monitoring:
    type: cloud-formation
    regions:
      - eu-west-1
    stacks:
      - cmp-monitoring
    app: monitoring
    contentDirectory: cdk.out
    parameters:
      templateStagePaths:
        CODE: CmpMonitoringStackEUCode.template.json
[...]

PR #670, opened 2022-10-06
ci: Remove Riff-Raff configuration workaround

akash1810:
## What does this change?

During a CloudFormation deployment, Riff-Raff sets the stack policy to protect against unintentional deletion of sensitive resources (see https://github.com/guardian/riff-raff/pull/623). The policy it uses is the same regardless of region. However, when the policy is applied to the us-west-1 region, we get an error.

We previously set `manageStackPolicy` to `false` to opt-out of this behaviour, as deployments to us-west-1 would otherwise be impossible.

https://github.com/guardian/riff-raff/pull/874 patches Riff-Raff to make the policy region aware, removing the need for this workaround.

## Why?
- More standard deployment configuration
- More protection during deployment
- Unblocks https://github.com/guardian/consent-management-platform/pull/659

## Notes
To be merged after https://github.com/guardian/riff-raff/pull/874.

### DIFF
--- a/riff-raff.yaml
+++ b/riff-raff.yaml
@@ -14,7 +14,6 @@ deployments:
                 CODE: CmpMonitoringStackEUCode.template.json
                 PROD: CmpMonitoringStackEUProd.template.json
             cloudFormationStackByTags: false
-            manageStackPolicy: false
     cmp-monitoring-lambda-eu-west-1:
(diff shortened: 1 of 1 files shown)


akash1810:
https://github.com/guardian/riff-raff/pull/874 is now shipped, making this safe to merge.

gnm-notona (approved):

PR #682, opened 2022-10-25
refactor: Simplify deploying to gh pages

sookburt:
## What does this change?

Simplify deployment to GH pages by using [gh-pages](https://www.npmjs.com/package/gh-pages) module to remove reliance on personal access tokens.

## Why?

Reducing risk by moving away from PAT (classic) a default token provided by GitHub.

co-authored-by with @akash1810


akash1810 (approved):

akash1810:
It looks like consent-management-platform is the [only user of guardian/ghpages](https://github.com/search?q=org%3Aguardian+guardian%2Fghpages&type=code), so we can archive that repository once this is merged too 🎉 .

PR #684, opened 2022-11-02
test: Monitor CMP Loading Time

akinsola-guardian:
## What does this change?
This PR monitors and logs the time to load the time to load a cmp banner in all locations.
## Why?
There have been reports of long loading times in Australia. Hence, it's useful to measure/calculate the average time to compare regions.


gnm-notona (commented):
Do we routinely use Cloudwatch metrics on the front end? 
I'm guessing there will be millions of these logged.  Are there any cost implications?

(Update: I may have misunderstood.  Are the metrics in a test process or in the production page?)

gnm-notona (approved):

PR #716, opened 2023-02-27
test: testing

sookburt:
This is an almost empty PR to fix the failing tests by using a fresh clone of the repository.

## Why?

Broken test in last PR merge


shtukas (approved):
For some strange reasons, the problem with the tests wasn't the tests, but the fact that there is something in Sue's old repository that made its way to the  main branch. We fixed it by closing the repo and we need this small PR to merge the repair to `main`

PR #729, opened 2023-03-27
fix!: Upgrade Typescript ~4.9.5 & Cypress

arelra:
## What does this change?

Upgrade:

- typescript@~4.9.5 to align with other repos
- cypress@^12.8.1

Cypress 12 now enforces [test isolation](https://www.cypress.io/blog/2022/12/06/announcing-cypress-12/#test-isolation-is-now-default). So I've also updated the e2e tests to not rely on previous test state and each test run initialises itself.

## Why?

Upgrade all the things

I attempted to release #724 previously with a `chore!` but it did not trigger a major release and it is blocking https://github.com/guardian/frontend/pull/25892.

I'm overriding the semantic prefix here with `fix!` to force a major release.

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -5,7 +5,10 @@ export default defineConfig({
   viewportHeight: 768,
   video: false,
   chromeWebSecurity: false,
-  retries: 2,
+  retries: {
+    runMode: 2,
+    openMode: 0,
+  },
   defaultCommandTimeout: 5000,
   e2e: {
     // We've imported your old cypress plugins here.
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -1,21 +1,21 @@
-import { defineConfig } from 'cypress'
+import { defineConfig } from 'cypress';
 
 export default defineConfig({
-  viewportWidth: 1440,
-  viewportHeight: 768,
-  video: false,
-  chromeWebSecurity: false,
-  retries: {
-    runMode: 2,
-    openMode: 0,
-  },
-  defaultCommandTimeout: 5000,
-  e2e: {
-    // We've imported your old cypress plugins here.
-    // You may want to clean this up later by importing these.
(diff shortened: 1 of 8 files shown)


cypress.config.ts
```diff
@@ -1,18 +1,21 @@
...
-    baseUrl: 'http://localhost:10001/',
-  },
-})
+	viewportWidth: 1440,
+	viewportHeight: 768,
+	video: false,
+	chromeWebSecurity: false,
+	retries: {
+		runMode: 2,
+		openMode: 0,
```
arelra, line 10: Just sets retries to 0 for open mode

cypress/e2e/sourcepoint-tcfv2.cy.js
```diff
@@ -107,16 +124,16 @@ describe('Interaction', () => {
...
 		cy.get(`[data-purpose="1"]`)
 			.should('have.data', 'consent')
-			.should('equal', false);
+			.should('equal', true);
 
 		[2, 3, 4, 5, 6, 7, 8, 9, 10].forEach((purpose) => {
 			cy.get(`[data-purpose="${purpose}"]`)
 				.should('have.data', 'consent')
-				.should('equal', true);
+				.should('equal', false);
```
arelra, line 136: I think this was a bug in the test but I could be wrong. Please do check

cypress/e2e/sourcepoint-aus.cy.js
```diff
@@ -21,11 +20,13 @@ const personalisedAdvertisingIs = (boolean) => {
 };
 
 describe('Window', () => {
-	loadPage(url);
 	it('has the guCmpHotFix object', () => {
+		cy.visit(url);
```
arelra, line 24: Each test now loads the url and initialises the browser state as required

PR #738, opened 2023-04-26
chore(deps): bump JSON5 to v2.2.2 via a resolution

sookburt:
## What does this change?

This PR adds a resolution to the package.json files to bump JSON5 to version 2.2.2

## Why?

JSON5 has a high vulnerability and is a transitive dependency on Jest which still exists in the latest version - therefore the need for a resolution.

Once Jest has a fixed transitive dependency, we can upgrade Jest as appropriate.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -26,7 +26,7 @@ jobs:
         run: yarn --frozen-lockfile
 
       - name: Run unit tests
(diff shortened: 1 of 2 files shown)


sookburt:
[Update JSON5 in CMP to fix High Dependabot Alert](https://trello.com/c/f6UkPZ3M/1021-update-json5-in-cmp-to-fix-high-dependabot-alert)

akinsola-guardian (approved):

PR #739, opened 2023-04-26
fix: remove fb vendor from vendor list

sookburt:
## What does this change?

Removed the Facebook vendor from the vendor list

## Why?

It is being removed from our IAB list and therefore must be removed from this.


akinsola-guardian (approved):

sookburt:
[Remove fb vendor from vendors.ts](https://trello.com/c/lDFXFfRU/1020-remove-fb-vendor-from-vendorsts)

PR #747, opened 2023-05-18
feat: upgrade node to 18.16.0

vlkerag:
## What does this change?

Upgrade to Node 18.6.0, upgrade to typescript 5.1.3, upgrade all dependencies as far as possible

## Why?

Security, maintenance

## Why are some dependencies not be upgraded?
- Inquirer: Version 9: Inquirer is now a native Node ECMAScript module. This will require your Node runtime to support es modules, and your app to be an es module https://github.com/SBoudrias/Inquirer.js/releases/tag/inquirer%409.0.0
- Rollup: Version 3: If you are using an ES module as configuration file, i.e. import and export syntax, you need to make sure Node will be loading your configuration as an ES module.  https://rollupjs.org/migration/
- monitoring: @tsconfig/node18: Need es2022 or lower but a higher version will use es2023
- cdk: aws-cdk has to matcht the one in @guardian/cdk#aws-cdk@2.78.0


vlkerag:
Merging into the node upgrade

.github/workflows/PR.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     name: Validate
     runs-on: ubuntu-latest
     steps:
-      - uses: amannn/action-semantic-pull-request@v5.1.0
+      -  uses: amannn/action-semantic-pull-request@v5.2.0
```
Mark-McCracken, line 14: I think this space needs to be removed
```suggestion
      - uses: amannn/action-semantic-pull-request@v5.2.0
```

monitoring/src/check-page/common-functions.ts
```diff
@@ -57,11 +61,12 @@ export const clearLocalStorage = async (page: Page): Promise<void> => {
...
 ): Promise<CustomPuppeteerOptions> => {
+	log_info(`chromium executable path: ${(process.env.IS_LOCAL == 'true') ? '/opt/homebrew/bin/chromium' : await chromium.executablePath()}`)
 	return {
 		headless: !isDebugMode,
-		args: isDebugMode ? ['--window-size=1920,1080'] : Chromium.args,
-		defaultViewport: Chromium.defaultViewport,
-		executablePath: await Chromium.executablePath,
+		args: isDebugMode ? ['--window-size=1920,1080'] : chromium.args,
+		defaultViewport: chromium.defaultViewport,
+		executablePath: (process.env.IS_LOCAL == 'true') ? '/opt/homebrew/bin/chromium' : await chromium.executablePath(),
```
Mark-McCracken, line 69: This might be a fix for the issue you're facing, but I'm not at all certain:
```suggestion
		executablePath: (process.env.IS_LOCAL == 'true') ? '/opt/homebrew/bin/chromium' : await chromium.executablePath('/opt/nodejs/node_modules/@sparticuz/chromium/bin'),
```
I found this from [this issue](https://github.com/Sparticuz/chromium/issues/41#issuecomment-1407545726)

.github/workflows/PR.yml
vlkerag: Thank you!

cdk/lib/monitoring.ts
```diff
@@ -38,8 +38,8 @@ export class Monitoring extends GuStack {
 				functionName: `${lambdaBaseName}-${stage}`,
 				fileName: `${lambdaBaseName}-lambda-${region}.zip`,
 				handler: 'index.handler',
-				runtime: Runtime.NODEJS_14_X,
-				timeout: Duration.seconds(300),
+				runtime: Runtime.NODEJS_18_X,
+				timeout: Duration.seconds(500),
```
akinsola-guardian, line 42: Why was the timeout updated?

vlkerag: Just for testing the lambda. Would expect to not increase them in the end

vlkerag:
Tested locally all yarn test, yarn validate, yarn audit, yarn lint, yarn start and all looking good for cmp, monitoring, cdk. Also tested the monitoring lambdas in code in all regions and appear to be consistently passing.  Also tested with yarn links and overrides for commercial/dotcom-rendering/frontend which looks like prod.

shtukas (approved):
Amazing work, 🎉 and admirable persistence.
Extra `console.log` in `src/onConsentChange.ts`

sookburt (approved):
Looks good! I've tested everything I can other than deploying it.

package.json
```diff
@@ -39,70 +39,60 @@
...
+    "semantic-release": "~21.0.5",
+    "serve": "^14.2.0",
+    "start-server-and-test": "~2.0.0",
+    "svelte": "~3.59.1",
+    "tslib": "^2.5.3",
+    "typescript": "~5.1.3",
     "wait-for-expect": "^3"
   },
   "peerDependencies": {
     "@guardian/libs": "^14.0.0"
```
joecowton1, line 96: This should be 15
```suggestion
    "@guardian/libs": "^15.0.0"
```

PR #748, opened 2023-05-19
test: fix broken monitoring after SP upgrade

sookburt:
## What does this change?

This change clears local storage as well as cookies before testing the second layer in cmp monitoring. 

## Why?

Following an upgrade to our underlying CMP to 4.9.0, consent is lingering in local storage from the previous test of the first layer and clearing cookies only no longer clears the consent ready to test the second layer.


vlkerag (approved):

sookburt:
[Fix TCFV2 Second Layer Monitoring following SP 4.9.0 upgrade](https://trello.com/c/HTnZzS3O/1038-fix-tcfv2-second-layer-monitoring-following-sp-490-upgrade)

PR #751, opened 2023-05-23
feat: version upgrades

vlkerag:


package.json
```diff
@@ -39,70 +39,70 @@
...
   },
   "peerDependencies": {
     "@guardian/libs": "^14.0.0"
   },
-	"resolutions":{
-		"//": "Once Jest upgrades its transitive dependency on json5 >=2.0.0, <2.2.2, remove this resolution",
-		"json5": "^2.2.2"
-	}
+  "dependencies": {
+    "jest-environment-jsdom": "^29.5.0"
```
akinsola-guardian, line 106: Should this be a dev dependency?

vlkerag: Yes agree

PR #760, opened 2023-06-08
chore: Update Typescript to v5.1.3 and tslib to v2.5.3

joecowton1:
## What does this change?

Updates TS and tslib versions as well as removing unused ESlint plugins and replacing with standard Guardian eslint config.

There is plenty more work to be done in this area to standardise linting in the project, but in terms of being able to update TS to version 5 without any dependency conflicts this works for now.

## Why?

We're updating TS to version 5.1.3 these across all Guardian packages


package.json
```diff
@@ -94,8 +94,8 @@
     "serve": "^11",
     "start-server-and-test": "~1.14.0",
     "svelte": "~3.52.0",
-    "tslib": "^2.4.1",
-    "typescript": "~4.9.5",
+    "tslib": "^2.5.3",
+    "typescript": "~5.1.3",
```
sndrs, line 98: appreciate it's possible outside the scope of this PR, but I think `tslib` should also be listed as a `peerDep`, because all non-relative imports are treated as external in the rollup config https://github.com/guardian/consent-management-platform/blob/main/src/rollup.config.js#L13

PR #765, opened 2023-06-15
fix: handle missing requestIdleCallback in vendorDataMananger

Jakeii:
## What does this change?
[`requestIdleCallback` is not available in Safari](https://caniuse.com/requestidlecallback) (!), we need to handle the case that it's not there.

Shoutout to @emma-imber for noticing this in sentry!


akinsola-guardian (approved):
Seems good!

PR #768, opened 2023-06-16
feat: promise.all

vlkerag:
## What does this change?
Monitoring:
- Wrap puppeteer into try catch finally to avoid memory leak in case of failure
- Clean up promise.all to not include an inner await or page.load/page.close

## Why?
- to avoid "Navigation failed because browser has disconnected!" or similar error.

### DIFF
--- a/cdk/lib/monitoring.ts
+++ b/cdk/lib/monitoring.ts
@@ -71,7 +71,7 @@ export class Monitoring extends GuStack {
 		});
 
 		const monitoringDuration: Duration =
-			stage === 'PROD' ? Duration.minutes(2) : Duration.days(1); // Every day for CODE; Every 2 minutes for PROD.
+			stage === 'PROD' ? Duration.minutes(2) : Duration.minutes(2); // Every day for CODE; Every 2 minutes for PROD.
 
 		new Rule(this, 'cmp monitoring schedule', {
 			schedule: Schedule.rate(monitoringDuration),
--- a/cdk/lib/monitoring.ts
+++ b/cdk/lib/monitoring.ts
@@ -71,7 +71,7 @@ export class Monitoring extends GuStack {
 		});
 
 		const monitoringDuration: Duration =
-			stage === 'PROD' ? Duration.minutes(2) : Duration.minutes(2); // Every day for CODE; Every 2 minutes for PROD.
+			stage === 'PROD' ? Duration.minutes(2) : Duration.days(1); // Every day for CODE; Every 2 minutes for PROD.
 
 		new Rule(this, 'cmp monitoring schedule', {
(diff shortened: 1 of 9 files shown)


vlkerag:
working in code

cdk/lib/monitoring.ts
```diff
@@ -71,7 +71,7 @@ export class Monitoring extends GuStack {
 		});
 
 		const monitoringDuration: Duration =
-			stage === 'PROD' ? Duration.minutes(2) : Duration.days(1); // Every day for CODE; Every 2 minutes for PROD.
+			stage === 'PROD' ? Duration.minutes(2) : Duration.minutes(2); // Every day for CODE; Every 2 minutes for PROD.
```
kenoir, line 74: Presumably this is for testing? Does it need to be switched back to a lower duration before merge?

kenoir (dismissed):
Looks good, one comment on the frequency in CODE that probably needs addressing.

monitoring/src/check-page/tcfv2.ts
```diff
@@ -133,12 +137,15 @@ const checkPages = async (config: Config, url: string, nextUrl: string) => {
 	await checkCMPLoadingTime(page, config);
 
 	await page.close();
-
+} catch (e) {
```
kenoir, line 140: I think the formatting here is a little funky, I don't think this project has a "format" task to run prettier.

kenoir:
Worth noting that some of the test logic here is duplicated (intentionally) in https://github.com/guardian/commercial-canaries/tree/main, should any of these changes also be in that repo?

kenoir:
> ## What does this change?
> Wrap page test in promise.all
> 
> ## Why?
> to avoid "Navigation failed because browser has disconnected!"

It's a bit difficult to tell the what/why from outside the team without more context, a bit more information would be helpful.

cdk/lib/monitoring.ts
vlkerag: Yes, it is only for testing and will be reverted before prod

vlkerag:
[Upgrade to node 18, upgrade to ts 5.1.3, upgrade as many dependencies as possible](https://trello.com/c/N042RtTX/1042-upgrade-to-node-18-upgrade-to-ts-513-upgrade-as-many-dependencies-as-possible)

vlkerag:
[Upgrade to node 18, upgrade to ts 5.1.3, upgrade as many dependencies as possible](https://trello.com/c/N042RtTX/1042-upgrade-to-node-18-upgrade-to-ts-513-upgrade-as-many-dependencies-as-possible)

sookburt (approved):

PR #775, opened 2023-06-20
fix: remove linkedin

vlkerag:
Remove linkedin from vendor list

## What does this change?
Remove linkedin

## Why?
Because linkedin was removed as a vendor


akinsola-guardian (dismissed):

sookburt (changes requested):
We also need to remove it from the vendors list in the README around line 329 please.

PR #777, opened 2023-06-23
test: remove promise.all

vlkerag:

### DIFF
--- a/cdk/lib/monitoring.ts
+++ b/cdk/lib/monitoring.ts
@@ -71,7 +71,7 @@ export class Monitoring extends GuStack {
 		});
 
 		const monitoringDuration: Duration =
-			stage === 'PROD' ? Duration.minutes(2) : Duration.days(1); // Every day for CODE; Every 2 minutes for PROD.
+			stage === 'PROD' ? Duration.minutes(2) : Duration.minutes(2); // Every day for CODE; Every 2 minutes for PROD.
 
 		new Rule(this, 'cmp monitoring schedule', {
 			schedule: Schedule.rate(monitoringDuration),
--- a/monitoring/src/check-page/aus.ts
+++ b/monitoring/src/check-page/aus.ts
@@ -34,10 +34,10 @@ const checkSubsequentPage = async (browser: Browser, url: string) => {
 	log_info(`Start checking subsequent Page URL: ${url}`);
 	const page: Page = await browser.newPage();
 	await loadPage(page, url);
-	await Promise.all([
-		checkCMPIsNotVisible(page),
-		checkTopAdHasLoaded(page),
-	]);
+	//await Promise.all([
+		await checkCMPIsNotVisible(page);//,
+		await checkTopAdHasLoaded(page);//,
+	//]);
 	await page.close();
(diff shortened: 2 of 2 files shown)


monitoring/src/check-page/aus.ts
```diff
@@ -19,7 +19,7 @@ const clickAcceptAllCookies = async (config: Config, page: Page) => {
 	log_info(`Clicking on "Continue" on CMP`);
 
 	const frame = await getFrame(page, config.iframeDomain);
-	await frame.waitForSelector(ELEMENT_ID.TCFV2_FIRST_LAYER_ACCEPT_ALL);
+	await frame.waitForSelector(ELEMENT_ID.TCFV2_FIRST_LAYER_ACCEPT_ALL), {visible: true};
```
akinsola-guardian, line 22: Has this been tested in CODE? Otherwise, seems okay.

sookburt (commented):
When running the tests in /monitoring, am consistently getting a timeout failure in common-functions.test.ts when it tries to find a frame. It seems to hit it remarkably fast for a timeout error though.  The weird thing is that it still seems to pass.  Something seems a bit off about it.

cdk/lib/monitoring.ts
```diff
@@ -71,7 +71,7 @@ export class Monitoring extends GuStack {
 		});
 
 		const monitoringDuration: Duration =
-			stage === 'PROD' ? Duration.minutes(2) : Duration.days(1); // Every day for CODE; Every 2 minutes for PROD.
+			stage === 'PROD' ? Duration.minutes(2) : Duration.minutes(2); // Every day for CODE; Every 2 minutes for PROD.
```
sookburt, line 74: Was it intentional to run CODE every 2 minutes also?

monitoring/src/check-page/tcfv2.ts
```diff
@@ -95,12 +77,12 @@ const checkSubsequentPage = async (
...
 		await checkTopAdHasLoaded(page);
 	await Promise.all([
-		clearCookies(await page.target().createCDPSession()),
+			clearCookies(page),
 		clearLocalStorage(page)
 	]);
 	await reloadPage(page);
 	await checkTopAdDidNotLoad(page);
-	await clickAcceptAllCookies(config, page);
+		await clickAcceptAllCookies(config, page, `Yes I'm Happy`);
```
sookburt, line 85: The reason we moved away from using the button text to identify and click on a button was because this often changes as part of a cookie banner test.  Then we get intermittent failures which are tricky to identify.

sookburt:
I tried running monitoring locally (using a VPN to mimic being in AUS and US).  TCFV2 in UK worked fine but AUS and US both failed consistently.  
On deploying to CODE and testing in the regions, they were inconsistent - with both failures and successes.

monitoring/src/check-page/tcfv2.ts
vlkerag: This is only for logging, same as it is at the moment. I only moved the function to a central place as it was duplicated, no functional changes.

PR #787, opened 2023-07-10
test: Create daily-test

akinsola-guardian:
## What does this change?
Run a scheduled test to compare the vendors list.

## Why?
We only notice if a vendor listed in vendors.ts list has been removed once we test locally. This PR will pick up any vendor changes.


Mark-McCracken (approved):
LGTM

PR #806, opened 2023-08-18
chore: Enable GPP Multi-State Privacy String (MSPS) with U.S. Privacy

akinsola-guardian:
## What does this change?
IAB is deprecating US privacy string - at the end of September and moving to the multi-state privacy agreement.

Google has said they will continue to support US Privacy String so not all vendors have signed up to use or accept the multi-state privacy string and have said they will continue to use the US privacy string.

To ensure full monetisation:

SP have a back up solution, Combination of the two (GPP and US Privacy String), so we will still be able to monetise those vendors for both frameworks

We need to implement BOTH:

MSPS (Multi-State Privacy String; GPP, Global Privacy Platform)

&

US privacy string (Current, soon to be legacy but still being supported)

https://trello.com/c/6qAHXGHq

https://docs.sourcepoint.com/hc/en-us/articles/18007731422099-Enable-GPP-Multi-State-Privacy-String-MSPS-with-U-S-Privacy-CCPA-solution

## Why?

Sourcepoint recommended change


kenoir (approved):
Tricksy! Looks good to me.

src/types/window.d.ts
```diff
@@ -97,5 +98,9 @@ declare global {
 			callback: (tcData: TCData, success: boolean) => void,
 			vendorIDs?: number[],
 		) => void;
+		__gpp?: (
+			command: string,
+			callback: (gppData: any, success: boolean) => void,
```
sookburt, line 103: should this be an 'any' type?

sookburt (approved):
We did some extra testing of the resulting GPP string and that the Privacy Manager appears, it's looking good! 🎉

PR #811, opened 2023-08-31
test: playwright

vlkerag:

### DIFF
--- a/monitoring/src/check-page/aus.ts
+++ b/monitoring/src/check-page/aus.ts
@@ -1,13 +1,12 @@
-import type { Browser, Page } from 'puppeteer-core';
+import type { Browser, BrowserContext, Page } from 'playwright-core';
 import type { Config } from '../types';
 import { ELEMENT_ID } from '../types';
 import {
 	checkCMPIsNotVisible,
 	checkCMPIsOnPage,
-	checkCMPLoadingTime,
+	//checkCMPLoadingTime,
 	checkTopAdHasLoaded,
 	clearCookies,
-	getFrame,
 	loadPage,
 	log_info,
 	makeNewBrowser,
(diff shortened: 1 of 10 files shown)


kenoir (commented):
Happy to see these getting some love 👍

monitoring/src/check-page/aus.ts
```diff
@@ -51,35 +39,32 @@ const checkPages = async (config: Config, url: string, nextUrl: string) => {
...
+	if (nextUrl) {
+		await checkSubsequentPage(context, nextUrl);
+	}
 
-		await checkCMPLoadingTime(page, config);
+	await page.close();
+	await browser.close();
 
-		await page.close();
+	//instead of clearing cookies and local storage, use a new browser and context, just using a new context did not work on lambda
```
akinsola-guardian, line 60: Can't we just reassign browser (if we use let)

akinsola-guardian: Same for page and context

monitoring/README.md
```diff
@@ -33,29 +33,24 @@ This CLI tool falls back to an interactive tool if any of the parameters passed
...
-To test locally you will need Chromium installed:
-
-`brew reinstall --cask chromium --no-quarantine`
+If you get the following error
 
-You also need to set the environment variable `export IS_LOCAL=true`.
+║ Looks like Playwright Test or Playwright was just installed or updated. ║
+║ Please run the following command to download new browsers:              ║
+║                                                                         ║
+║     yarn playwright install                                             ║
```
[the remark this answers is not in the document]
vlkerag: Unfortunately 'yarn playwright install' does not work which is why I added this explanation of what to do instead.

monitoring/src/check-page/aus.ts
vlkerag: Yes, will give it more meaningful variable names in a cleanup. Thanks!

sookburt (approved):
On testing locally with a VPN, the ccpa monitoring (only) fails frequently but on CODE it is succeeding fine.  We may need to allow for this when running locally with a VPN.

akinsola-guardian (approved):

PR #813, opened 2023-09-05
feat: Remove Lotame from vendors.ts

SiAdcock:
## What does this change?

Removes Lotame from vendors.ts

## Why?

We deactivated Lotame [a while back](https://github.com/guardian/commercial/pull/138), so consent checks for this vendor are no longer needed.

Tangential: we are seeing [a large number of errors in Sentry](https://the-guardian.sentry.io/issues/4355550636/?project=1377847&query=is%3Aunresolved&referrer=issue-stream&sort=freq&statsPeriod=14d&stream_index=1) regarding failing calls to Lotame's server. Removing Lotame from the vendor list will isolate the CMP as a potential cause of these requests.


sookburt (approved):
Thanks @SiAdcock - we're about to merge another similar change so if you want to merge, we will merge our change too and we can test and update all the consumers with the joint changes.

sookburt (approved):
Merged changes from last patch - looks good

PR #840, opened 2023-11-01
A first iteration of launch.json file for debugging monitoring

sookburt:
## What does this change?

No change to code. This works to debug monitoring against prod with 3 different options (tcfv2, aus and ccpa).
It can be altered or extended to run against code or local if needed

## Why?

In case others in the team find this useful.


sookburt:
It may not be relevant once CMP is moved to libs though...

vlkerag (approved):

PR #845, opened 2023-11-09
chore: Add @guardian/eslint-config and update eslint

akinsola-guardian:
## What does this change?
As recommended by Joe Cowton's PR, I've installed @guardian/eslint-config' and using it with '@guardian/eslint-config-typescript' in eslintrc.js throughout the app
## Why?
To ensure that CMP linting standards align with the patterns recommended Client Side


cdk/.eslintignore
```diff
@@ -2,3 +2,5 @@ node_modules
 cdk.out
 .eslintrc.js
 jest.config.js
+**.d.ts
```
akinsola-guardian, line 5: Check if this is the best way to respond to this error:

 0:0  error  Parsing error: ESLint was configured to run on `<tsconfigRootDir>/bin/cdk.d.ts` using `parserOptions.project`: /users/akinsola_lawanson/code/consent-management-platform/cdk/tsconfig.json
However, that TSConfig does not include this file. Either:
- Change ESLint's list of included files to not include this file
- Change that TSConfig to include this file
- Create a new TSConfig that includes this file and include it in your parserOptions.project
See the typescript-eslint docs for more info: https://typescript-eslint.io/linting/troubleshooting#i-get-errors-telling-me-eslint-was-configured-to-run--however-that-tsconfig-does-not--none-of-those-tsconfigs-include-this-file

src/aus/api.test.js
```diff
@@ -1,4 +1,4 @@
-import { getUSPData } from './api';
+import { getUSPData } from './api.ts';
```
joecowton1, line 1: Do we understand why these `.ts` have to be added? I think i did the same on my original PR for these changes but looking at other repos that use the eslint config it doesn't always seem necessary.

akinsola-guardian: I'm not sure. It seems to be because it's a js file importing a ts file. It also relates to this rule (https://github.com/import-js/eslint-plugin-import/blob/v2.27.5/docs/rules/no-unresolved.md)

.eslintrc.js
```diff
@@ -4,18 +4,29 @@ module.exports = {
 		jest: true,
 		browser: true,
 	},
-	extends: ['@guardian/eslint-config-typescript'],
```
mxdvl, line 7: I think we still want to extend from the base config, [as per the docs](https://github.com/guardian/csnx/tree/main/libs/@guardian/eslint-config-typescript#usage).

cdk/.eslintignore
mxdvl: I think that you would want to edit CDK’s `tsconfig.json` instead.

cdk/package.json
```diff
@@ -17,9 +17,10 @@
 		"diff": "cdk diff --path-metadata false --version-reporting false"
 	},
 	"devDependencies": {
+		"@guardian/eslint-config": "^5.0.0",
```
mxdvl, line 20: This is redundant with `@guardian/eslint-typescript-config`, which includes the javascript one automatically. We do not need this.

monitoring/src/check-page/common-functions.ts
```diff
@@ -360,7 +363,7 @@ export const sendMetricData = async (
 	log_info(`config.platform.toUpperCase() ${config.platform.toUpperCase()})`);
 	const region = config.region;
 	const client = new CloudWatchClient({ region: region });
-	const params = {
+	const params: PutMetricDataCommandInput = {
```
mxdvl, line 366: You can also use [the `satisfies` operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator) here if you want a narrower type.

monitoring/tsconfig.eslint.json
```diff
@@ -1,5 +1,5 @@
 {
 	"extends": "./tsconfig.json",
-	"include": ["src/**/*", "*.ts", "node_modules/puppeteer-core/**/*"],
+	"include": ["src/**/*", "*.ts", "node_modules/playwright-core/**/*"],
```
mxdvl, line 3: Why are we type-checking playwright?

monitoring/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
 	"extends": "@guardian/tsconfig",
-	"include": ["./*.ts"],
+	"include": ["./*.ts", "./**/*.ts"],
```
mxdvl, line 3: Slightly redundant:

```suggestion
	"include": ["./**/*.ts"],
```

package.json
```diff
@@ -47,7 +47,7 @@
 		"@babel/preset-typescript": "^7.23.2",
 		"@babel/runtime": "^7.23.2",
 		"@changesets/cli": "^2.26.2",
-		"@guardian/eslint-config": "^4.0.0",
+		"@guardian/eslint-config": "^5.0.0",
 		"@guardian/eslint-config-typescript": "^6.0.0",
```
mxdvl, line 51: As above, we only need one of these, and I’d recommend using `@guardian/eslint-config-typescript` v7

src/aus/api.test.js
mxdvl: Yes, this rule makes imports much more explicit 👍

mxdvl (commented):
I strongly believe that we should still extend from the base configs.

I would also recommend doing the formatting/Prettier changes as part of a different PR. The `package.json` diffs are adding unnecessary noise to this PR.

monitoring/src/check-page/common-functions.ts
akinsola-guardian: Thanks 👍

cdk/package.json
akinsola-guardian: Thanks. I removed this.

monitoring/tsconfig.eslint.json
akinsola-guardian: Just swapped it over but have removed it.

monitoring/tsconfig.json
akinsola-guardian: Thanks.

.eslintrc.js
akinsola-guardian: We're extending it for all the `'.ts','*.tsx'` and using @guardian/eslint-config for `'.js','*.jsx'` in the overrides.

cdk/.eslintignore
akinsola-guardian: Do you how I would do this? I tried adding it to the exclude but that didn't work.

.eslintrc.js
mxdvl: Why is that?

akinsola-guardian: To make sure it's only applied to .ts files. When i use this to extend all files. I get this error:  

`
/code/consent-management-platform/jest.config.js
  4:13  error  Require statement not part of import statement  @typescript-eslint/no-var-requires
  `

akinsola-guardian: Thansk, I found a way to resolve this.

cdk/.eslintignore
akinsola-guardian: Thanks this has been resolved.

mxdvl (dismissed):
I think this looks good now!

sookburt (approved):

PR #863, opened 2023-12-12
cmpVersion as integer into metrics

vlkerag:
## What does this change?
Save the cmp version used on dotcom so that it can be monitored for debugging

## Why?
To more accurately pin down Sourcepoint changes for debugging purposes.


akinsola-guardian (approved):
LGTM!

PR #868, opened 2023-12-18
default consents for 11 is false

vlkerag:
## What does this change?
Default consent 11 to false. Note that this will set canTarget to false if only purpose 11 is de-selected which is not currently the case.

## Why?
Consent 11 is currently 'undefined' if reader rejected all or explicitly rejected purpose 11


akinsola-guardian (approved):
Seems good. Tested this locally using yarn test and yarn start.

akinsola-guardian:
This was implemented in the csnx repo

akinsola-guardian:
https://github.com/guardian/csnx/pull/1391

PR #880, opened 2023-12-29
chore: Create group for only minor and patch version updates

akinsola-guardian:
## What does this change?
This creates a dependabot group PR release only for minor and patch version updates.
## Why?
I previously merged a PR that grouped updates based on the directory. which created three failing PR's with 12,14,40 updates. These are quite difficult to resolve as there many packages being updated at once. 

This PR will pick up changes that don't break their API/Contract so they'll be less likely to fail.
## Link to Trello
https://trello.com/c/ntimUWJd


sookburt:
Is it worth adding a Day, say Sunday, so that we have a set of fresh PRs ready to go on Monday?

sookburt:
I notice the link to the documentation is out of date - perhaps we could update that while we're here?  I assume it's this one: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

sookburt:
What do we want to do about Typescript bumps since it doesn't follow semver in the traditional way?

PR #895, opened 2024-01-15
Test moving monitoring tests into CMP for consumer tests with tree shaking

sookburt:
This PR is just testing what is possible for testing the consumers of CMP when we create a beta release.

We are hoping to find a way to reduce the time spent regression testing the consumers against our changes to the CMP before we release a new library.

The idea is that each consumer has a GHA that tests the application against a narrow subset of the same tests we use in monitoring.

Each consuming application loads CMP as a dev dependency (as well as a dependency), a script is added to the consuming repo to pass details to a GH Action that runs the CMP monitoring tests against its own URL and report back the result.

Each consumer will then tree shake to remove the testing code on release.

We can only use tree shaking if each consumer is using EMS.

### DIFF
--- a/monitoring/src/check-page/aus.ts
+++ b/monitoring/src/check-page/aus.ts
@@ -1,5 +1,4 @@
 import type { Browser, BrowserContext, Page } from 'playwright-core';
-import type { Config } from '../types';
 import {
 	checkCMPIsNotVisible,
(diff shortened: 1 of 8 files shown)


sookburt:
Closing for several reasons:
- Leaving the team

PR #903, opened 2024-01-26
Consumer self test investigation (2)

sookburt:
## What does this change?

Trying to get a subset of the monitoring tool into cmp to act as a cmp consumer self test.

## Why?

To automate the testing of our BETA releases in each of our consumers in conjunction with dependabot.  This is hoped to reduce the amount of manual regression testing that we're currently doing.

## Link to Trello
https://trello.com/c/hXMAuC6X


sookburt:
When I uncomment the export statement for consumer-self-test in index.ts, the test-page gives a 404 and the Cypress testing fails consistently.  I suspect a clash of global Window object between these.

sookburt:
Closing for several reasons:
- Leaving the team

PR #913, opened 2024-01-30
countries from @guardian/libs

vlkerag:
## What does this change?
Use countries from @guardian/libs

## Why?
Avoid code duplication and improve maintainability


sookburt:
A slight concern if the Åland Islands ever need their own framework as it does not match on name owing to the diacritic (unlikely and will currently get tcfv2 by default as getting the framework seems to be the only use of the countries) but other than that the two countries lists appear to be from the same set.

akinsola-guardian:
Seems good. Has this been tested in DCR? Noticed this comment about taking a string to integrate into DCR. https://github.com/guardian/consent-management-platform/blob/8f5e07ce109178d694b77f7995294fce6287b9f2/src/types/countries.ts#L257

PR #915, opened 2024-02-02
Revoke gpp ccpa stub

akinsola-guardian:
## What does this change?
Removes __gpp stub from loading in CCPA.
## Why?
This caused an issue with ads. This is a temporary removal and will be reinstated once it's been debugged.

Tested using about us to check the stubs loaded.

### DIFF
--- a/.changeset/spotty-kangaroos-change.md
+++ b/.changeset/spotty-kangaroos-change.md
@@ -1,0 +1,5 @@
+---
+"@guardian/consent-management-platform": minor
+---
+
+Removing \_\_gpp stub load in CCPA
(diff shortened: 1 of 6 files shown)


sookburt (commented):
Based on the earlier PR to implement GPP, there were also changes in the window.d.ts file that aren't appearing here: https://github.com/guardian/consent-management-platform/commit/60b7d1501572d84749df9b20299fa44cbdfa674d. Do they need to be removed too?

sookburt (commented):
Following our earlier discussion, if we're wanting to fully strip everything 'GPP' out, with a view to moving to the full switch over process in April without the intermediate solution (to avoid later confusion), I think there are some extra bits that need removing to be on the safe side:

1. at line 38 in window.d.ts, remove 'includeGppApi (and the import at line 6),
2. remove test at line 57 in sourcepoint.test.js
3. remove mention of gpp in comment ln 159 in sourcepoint.ts
4. ccpa.ts line 10 remove the interface for GPPData (?)
5. delete the stub_gpp_ccpa.js file (?),

sookburt (approved):
Thanks - looks good! 😃

PR #920, opened 2024-02-06
Consumer self test v3

sookburt:


sookburt:
Closing for several reasons:
- Leaving the team

PR #921, opened 2024-02-08
Updating prebid vendor

akinsola-guardian:
## What does this change?
Adds prebidCustom vendor.
## Why?
prebidCustom has been added as a custom vendor.


sookburt (approved):
LGTM

PR #950, opened 2024-03-21
add new comscore cookie to `vendorStorageIds`

Jakeii:
## Why
ComDev is making a change that will allow comscore to drop this first party cookie. https://github.com/guardian/commercial/pull/1313

By adding it here we can make sure it's removed if consent changes.


akinsola-guardian (approved):
LGTM!

PR #959, opened 2024-04-02
Monitoring reload print

vlkerag:
## What does this change?
Add more printing when re-load fails

## Why?
Debugging


vlkerag:
Not required

PR #989, opened 2024-05-16
Fix Monitoring by defining the NodeJS runtime version

akinsola-guardian:
## What does this change?
- Update playwright
- Fixing the runtime version to previous Lambda Runtime ARN.
## Why?
The automatic update is incompatible with playwright-aws-lambda

https://github.com/JupiterOne/playwright-aws-lambda/issues/74

Tested in CODE

### DIFF
--- a/cdk/lib/__snapshots__/monitoring.test.ts.snap
+++ b/cdk/lib/__snapshots__/monitoring.test.ts.snap
@@ -47,7 +47,7 @@ exports[`The Monitoring stack matches the snapshot 1`] = `
         },
         "Runtime": "nodejs18.x",
         "RuntimeManagementConfig": {
-          "RuntimeVersionArn": "0cdcfbdefbc5e7d3343f73c2e2dd3cba17d61dea0686b404502a0c9ce83931b9",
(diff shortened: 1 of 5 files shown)


cdk/lib/__snapshots__/monitoring.test.ts.snap
```diff
@@ -46,6 +46,10 @@ exports[`The Monitoring stack matches the snapshot 1`] = `
           ],
         },
         "Runtime": "nodejs18.x",
+        "RuntimeManagementConfig": {
+          "RuntimeVersionArn": "arn:aws:lambda:eu-west-1::runtime:0cdcfbdefbc5e7d3343f73c2e2dd3cba17d61dea0686b404502a0c9ce83931b9",
```
coldlink, line 50: we generally try and avoid putting ARNs in public projects, I'm guessing there's no way to make this a secret or environment variable?

i think it'll be fine though if we don't

coldlink (approved):

cdk/lib/__snapshots__/monitoring.test.ts.snap
akinsola-guardian: Thanks for this. I'm now using a github secret which removed it from the monitoring.ts file but this snapshot file that stores the template will always contain this. I don't know how to get around this.

coldlink (approved):

PR #995, opened 2024-05-24
Add runtime id to monitoring

akinsola-guardian:
## What does this change?
Adding the runtime id to the cdk. Removing the runtime id secret
## Why?
In this PR (https://github.com/guardian/consent-management-platform/pull/989), I added the lambda node js runtime id as a secret. This has made local testing more difficult without actually removing the id from the repo (monitoring cdk snapshot)


raphaelkabo (approved):
Looks good!

PR #1026, opened 2024-06-27
Refactor env vars

akinsola-guardian:


monitoring/src/index.ts
```diff
@@ -14,7 +14,7 @@ export const handler = async (
 	const stage: string = event.stage;
 	const region: AwsRegionOpt = event.region;
 
-	const configWrapper = new ConfigWrapper(region, stage, jurisdiction);
+	const configWrapper = new ConfigWrapper(region, stage, jurisdiction, stage );
```
vlkerag, line 17: What values could stage be here?

PR #1033, opened 2024-07-01
Re-implement AMP page checks.

akinsola-guardian:
## What does this change?
- Moving from using the `page.context().clearCookies()` to iteratively clear the list of cookies. 
- Created script `run.sh` to call `yarn start` repeatedly.
- Applying `CheckPagesProps` type to tcfv2 function.

## Why?
AMP pages were not clearing cookies effectively.
This has been tested locally and in CODE.


monitoring/src/check-page/common-functions.ts
```diff
@@ -47,7 +47,12 @@ export const ScreenDimensions = {
  * @return {*}  {Promise<void>}
  */
 export const clearCookies = async (page: Page): Promise<void> => {
-	await page.context().clearCookies();
+	const cookies = await page.context().cookies();
+
+	for(const cookie of cookies) {
+		await page.context().clearCookies(cookie);
```
AshCorr, line 53: How is this better than `clearCookies()`? Could maybe do with a comment for future travellers so nobody ends up "optimizing" this and reverting your change!

AshCorr (approved):

AshCorr:
AMP is now dead 🫡

PR #1043, opened 2024-07-15
upgrade versions, eslint, jest, node

vlkerag:
## What does this change?
Upgrade versions as much as possible:
To node 20
To newer eslint version which requires eslint.config.mjs
All other versions upgrade except for inquirer where a newer version would require esm.

The changes to localRun.ts and remoteRun.ts would be required for a newer inquirer version. While this PR does not increase the inquirer version yet, I would like to make those changes now to avoid them in the future and not lose the work I had done on it.

## Why?
Upkeep and some prep to move to esm in the future


cdk/package.json
```diff
@@ -11,30 +11,30 @@
...
+		"ts-jest": "^29.2.2",
 		"ts-node": "^10.9.2",
-		"tslib": "^2.6.2",
+		"tslib": "^2.6.3",
 		"typescript": "~5.5.3"
 	},
 	"dependencies": {
-		"npm": "^10.4.0",
-		"source-map-support": "^0.5.20"
+		"npm": "^10.8.2",
```
raphaelkabo, line 36: Do we need `npm` as a dependency if the `cdk` directory is using `pnpm`?

cdk/script/export-monitoring-file
```diff
@@ -19,7 +19,14 @@ then
 
 	echo "Zipping Built Files"
 	cd "${absolute_src_directory}" # Go to monitoring/dist folder
-	zip -FSr "${ROOT_DIR}/cdk/dist/cmp-monitoring-lambda.zip" "." # Zip the monitoring/dist folder into cdk/dist
+
+	newDir="${ROOT_DIR}/cdk/dist/"
+
+	if ! [[ -d "$newDir" ]] ; then
+		mkdir -p "$newDir"
+	fi
```
raphaelkabo, line 27: The if statement can be removed here because `mkdir -p` fails silently if the directory already exists:

> with this option specified, no error will be reported if a directory given as an operand already exists.

raphaelkabo:
Just a couple of comments, but this looks good overall - the only thing is it might be useful to leave a bit more information in the PR description to let future people know what needs to happen in the code to switch over to ESM.

cdk/script/export-monitoring-file
vlkerag: Thanks!

cdk/package.json
vlkerag: Removed

vlkerag:
Did run in code

PR #1045, opened 2024-07-17
newer inquirer versions and type-check

vlkerag:
## What does this change?
Use the newest inquirer version which actually means using "@inquirer/prompts". Inquirer 9 only supported esm so we could not upgrade from 8 but the new "@inquirer/prompts" supports cjs again. The new "@inquirer/prompts" also has a different syntax, hence using the new syntax.

## Why?
It will prevent future vulnerabilities in an old inquirer version. Inquirer changed to use @inquirer/prompts: https://github.com/SBoudrias/Inquirer.js/releases/tag/inquirer%4010.0.0

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -31,9 +31,6 @@ updates:
         update-types:
           - 'minor'
           - 'patch'
-        exclude-patterns:
-          - 'inquirer'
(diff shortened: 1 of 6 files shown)


akinsola-guardian (approved):
Thank you

PR #1119, opened 2024-12-03
Update node version to 22.12.0

akinsola-guardian:
## What does this change?
Update node version to 22.12.0

## Why?
Keeping version up to date.

Tested in CODE


paulmr (approved):

coldlink (changes requested):
I suggest using the LTS version rather than latest, as the odd version Node is only supported for 6 months

https://github.com/nodejs/release?tab=readme-ov-file#release-schedule

So I Suggest the latest 22.x version

coldlink (approved):

PR #1169, opened 2025-05-12
Add aws synthetics

akinsola-guardian:
## What does this change?

This PR adds CloudWatch Synthetics canaries to monitor the CMP banners. It's the first step in replacing our current monitoring app, which relies entirely on AWS Lambdas.

Our existing  monitoring required an update from Node.js 18  in June. However, due to limitations with the playwright-aws-lambda package used to support Playwright running on AWS Lambda — shown [here](https://github.com/JupiterOne/playwright-aws-lambda/issues/74) — we were locked to a specific runtime version. This restriction made upgrading difficult and unsustainable.

As a result, we’re migrating to AWS-supported Synthetics, which provide easy use of Playwright 

The synthetic script is written in JavaScript because the @amzn/synthetics-playwright library is not published to the public npm registry.

## Why?
To replace the older monitoring.


.github/workflows/monitoring-CI.yml
```diff
@@ -49,6 +49,36 @@ jobs:
...
+          package_json_file: "synthetics/package.json"
+
+      - name: Setup synthetics node
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: "./synthetics/.nvmrc"
+          cache: pnpm
+          cache-dependency-path: |
+            synthetics/pnpm-lock.yaml
+            cdk/pnpm-lock.yaml
```
AshCorr, line 64: Is Node happy with being setup twice? I'm guessing the cache might not be working correctly?

cdk/lib/monitoring.ts
```diff
@@ -104,23 +112,79 @@ export class Monitoring extends GuStack {
...
+			},
+			test: synthetics.Test.custom({
+				handler: 'index.handler',
+				code: synthetics.Code.fromBucket(
+					Bucket.fromBucketName(this, 'BucketName', bucketName),
+					`${stage}/${canaryBaseName}-${region}/nodejs.zip`,
+				),
+			}),
+			provisionedResourceCleanup: true,
+			runtime: synthetics.Runtime.SYNTHETICS_NODEJS_PLAYWRIGHT_1_0,
```
AshCorr, line 169: 🥳

cdk/script/export-synthetics-file
```diff
@@ -0,0 +1,31 @@
...
+}
+
+if [[ -f "${INDEX_JS_BUILT_FILE}" ]]; then
+  absolute_src_directory=$(get_absolute_path_directory "$INDEX_JS_BUILT_FILE")
+
+  echo "Zipping built files from: $absolute_src_directory"
+  mkdir -p "$DEST_DIR"
+  cp -r "${absolute_src_directory}/." "$DEST_DIR"
+
+  cd "${ROOT_DIR}/cdk/dist"
```
AshCorr, line 23:
```suggestion
  cd "${DEST_DIR}"
```

synthetics/src/main.js
```diff
@@ -0,0 +1,38 @@
...
+			} else if (error instanceof Error) {
+				errorMessage = error.message;
+			}
+
+			console.log(
+				`(cmp monitoring) Finished with failure: ${errorMessage}`,
+			);
+
+			throw new Error(errorMessage);
+		}
```
AshCorr, line 30: `catch` inside a `try` with a `catch` feels a little redundant!

AshCorr (dismissed):
LGTM apart from a minor comment or two! Assumed most of the code is what was already there for the previous lambdas just un-typescriptified so haven't scrutizined it too closely!

AshCorr (approved):

PR #1171, opened 2025-05-20
fix: Swap USNat for CCPA

akinsola-guardian:
## What does this change?
Using the correct value when checking the jurisdiction.
Remove optout check as it's inconsistent

## Why?
The name was changed from CCPA to USNAT throughout the codebase.
## Link to Trello

Tested locally.


paulmr:
Looks good to me!

paulmr (approved):

PR #1172, opened 2025-06-11
Create makefile

akinsola-guardian:
## What does this change?
Creates a makefile to make it easier to use repo.
## Why?
To make it easier to run functions in synthetics directory.


paulmr (approved):
looks great!

PR #1185, opened 2025-07-10
Fix errors which prevented synthetics-start-debug from running

codingWithRach:
## What does this change?

Fixes a number of errors which prevented `synthetics-start-debug` from running in my environment

## Why?

I have been unable to test the changes in [this](https://github.com/guardian/csnx/pull/2118) csnx PR

## Link to Trello

https://trello.com/c/zMn6jaUm

### DIFF
--- a/synthetics/src/main.js
+++ b/synthetics/src/main.js
@@ -12,6 +12,8 @@ export const main = async (browserType, region, stage) => {
 		Log.info("Config generated:", configWrapper.config);
 
 		await configWrapper.run(browserType);
+		
+		Log.info("Tests completed successfully");
 	} catch (error) {
 		let errorMessage = "Unknown Error!";
 
@@ -24,8 +26,5 @@ export const main = async (browserType, region, stage) => {
 		Log.error(`Finished with failure: ${errorMessage}`);
 		console.error("Error in handler:", error);
 		throw error; // Rethrow the error to ensure the Lambda function fails
-	} finally {
-		// Ensure browser is closed
-		await browserType.close();
 	}
 };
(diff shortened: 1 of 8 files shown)


synthetics/src/main.js
```diff
@@ -24,8 +26,5 @@ export const main = async (browserType, region, stage) => {
 		Log.error(`Finished with failure: ${errorMessage}`);
 		console.error("Error in handler:", error);
 		throw error; // Rethrow the error to ensure the Lambda function fails
-	} finally {
-		// Ensure browser is closed
-		await browserType.close();
```
rBangay, line 29: Hi .. just trying to understand why this was removed?

codingWithRach: I removed it because it caused an error - possibly because the browser has already been closed at that point.  Having removed it, there's no error, and at the end of `make synthetics-start-debug` no testing browser remains open

synthetics/src/utils/cmp-checks.js
```diff
@@ -48,8 +48,19 @@ export const checkCMPIsNotVisible = async (page) => {
...
+	} catch (error) {
+		if (error.name === 'TimeoutError') {
+			// If it times out, check one more time if it's visible
+			if (await cmpl.isVisible()) {
+				Log.error("CMP is still visible after timeout");
+				throw Error("CMP still present on page");
+			}
+		} else {
+			throw error;
+		}
```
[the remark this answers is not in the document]
rBangay: Thanks for explaining 👍  .. sorry I'm new to the CMP. Do you know what causes the delay in hiding the CMP? and do we need to wait 10 seconds for it?

synthetics/src/main.js
codingWithRach: It seems that browserType is actually a browser type object (synthetics or chromium) rather than a browser instance, so doesn't have an isConnected or close method

codingWithRach: I should have said that when I answered your original question about it really, but it's been a fortnight since I worked on this and I didn't remember the details!

PR #1191, opened 2025-07-30
Rachel/fix make commands

codingWithRach:
## What does this change?

Corrects make commands or scripts calls for:
- make synthetics-test-dev
- make cdk-build
- make cdk-test-dev
- make cdk-lint
- make cdk-diff

Adds missing commands:
- make cdk-test-update
- make install

Remove unneeded command:
- make cdk-watch

Updates list of commands in `make help`

## Why?

There were a number of errors when I ran all the commands in the makefile

## Link to Trello

https://trello.com/c/ZRUq2Ckz

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,8 +1,9 @@
-.PHONY: help clean
+.PHONY: help clean update install
 .PHONY: synthetics-install synthetics-test synthetics-test-dev synthetics-lint
 .PHONY: synthetics-build synthetics-start synthetics-clean synthetics-update
 .PHONY: synthetics-start-debug
-.PHONY: cdk-install cdk-build cdk-watch cdk-test cdk-test-dev cdk-format cdk-lint cdk-synth cdk-diff cdk-clean cdk-update
+.PHONY: cdk-install cdk-build cdk-test cdk-test-dev cdk-test-update cdk-format
+.PHONY: cdk-lint cdk-synth cdk-diff cdk-clean cdk-update
 # Makefile for managing synthetics and CDK projects
 # This Makefile provides a set of commands to manage the synthetics and CDK projects
 # including installation, testing, linting, building, and cleaning.
@@ -21,7 +22,6 @@
 #   synthetics-update: Update dependencies for the synthetics project
 #   cdk-install: Install dependencies for the CDK project
 #   cdk-build: Build the CDK project
-#   cdk-watch: Build the CDK project in watch mode
 #   cdk-test: Run tests for the CDK project
 #   cdk-test-dev: Run tests for the CDK project in watch mode
 #   cdk-test-update: Run tests for the CDK project and update snapshot files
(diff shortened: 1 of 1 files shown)


makefile
```diff
@@ -143,7 +150,12 @@ cdk-synth:
 
 cdk-diff:
 	$(call log, "Showing CDK diff...")
-	cd cdk && pnpm run diff
+	@if [ -z "$$AWS_PROFILE" ]; then \
+		echo "Error: AWS_PROFILE is not set - please run 'export AWS_PROFILE=<profile>' before running this command."; \
+    	exit 1; \
+	else \
+		cd cdk && pnpm run diff; \
```
AshCorr, line 157: Could we set the AWS_PROFILE here for the user as I take it this project is only deployed to the one AWS account?


```suggestion
		cd cdk && AWS_PROFILE=frontend pnpm run diff; \
```

cdk/package.json
```diff
@@ -1,45 +1,47 @@
...
+	"scripts": {
+		"build": "tsc -p tsconfig.build.json",
+		"test": "jest",
+		"test:dev": "jest --watch",
+		"test-update": "jest -u",
+		"format": "prettier --write \"{lib,bin}/**/*.ts\"",
+		"lint": "eslint \"lib/**/*.ts\" \"bin/**/*.ts\" --no-error-on-unmatched-pattern",
+		"synth": "cdk synth --path-metadata false --version-reporting false",
+		"diff": "cdk diff --path-metadata false --version-reporting false",
+		"watch": "tsc -p tsconfig.build.json --watch"
```
AshCorr, line 14: Do we need a `watch` script for CDK? Since CDK is usually a run and done kind of thing, theres not a server thats constantly running that needs to be reloaded.

cdk/package.json
```diff
@@ -1,45 +1,47 @@
...
-    "setupFilesAfterEnv": [
-      "./jest.setup.js"
-    ]
-  },
-  "packageManager": "pnpm@10.4.1+sha512.c753b6c3ad7afa13af388fa6d808035a008e30ea9993f58c6663e2bc5ff21679aa834db094987129aa4d488b86df57f7b634981b2f827cdcacc698cc0cfb88af"
+	"name": "cdk",
+	"version": "0.0.0",
+	"private": true,
+	"scripts": {
+		"build": "tsc -p tsconfig.build.json",
```
AshCorr, line 6: Curious why a new `tsconfig.build.json` file is required! The `tsc` command should default to using the existing `cdk/tsconfig.json` command if you don't specify another one!

codingWithRach: If I try replacing the existing tsconfig.json with the one I'm currently using for tsconfig.build.json, then all commands work EXCEPT the addition of the `outdir: dist` breaks cdk-synth and cdk-diff (but if I omit `outdir:dist` then cdk-lint complains)

AshCorr: Ah hm, yea it looks like a lot of CDK projects are broken in the same way. I guess people usually don't really build their CDK projects, they just run `npm synth`

codingWithRach: Well actually it seems this project already has 2 tsconfig files, because there's also a tsconfig.eslint.json...

AshCorr: Not that its something that needs to be addressed in this PR ofc

codingWithRach: I'll have a play around and see if we can get it down to 1

PR #1193, opened 2025-08-06
Add Sourcepoint environment in synthetics

akinsola-guardian:


codingWithRach (approved):
This is brilliant, and worked perfectly when I ran it with local/stage and prod/prod for Australia (but haven't yet been able to check whether it works for code/stage)

codingWithRach (changes requested):
Unfortunately `_sp_env=stage` doesn't get appended when running in CODE (which is the same problem I encountered when trying to do this manually, before the existence of this PR)

codingWithRach (approved):
I've checked local, code and prod for all jurisdictions and all tests pass.  Great work - thanks!

PR #1195, opened 2025-08-07
Ensure only one tsconfig file for cdk

codingWithRach:
## What does this change?

Ensures that `cdk` only has a single `tsconfig.json` file, necessitating a change to the following cdk scripts:
- build
- synth
- diff

## Why?

The project had ended up with three separate tsconfig files:
- tsconfig.json
- tsconfig.build.json
- tsconfig.eslint.json

The latter didn't appear to be in use, and the other 2 have been merged

## Link to Trello

https://trello.com/c/VnX6Y8Dy


cdk/tsconfig.json
```diff
@@ -1,12 +1,5 @@
 {
-  "extends": "@guardian/tsconfig/tsconfig.json",
-  "compilerOptions": {
-    "module": "CommonJS"
-  },
```
AshCorr, line 5: Is removing `"module": "CommonJS"` the reason why we need to specify file extensions in imports now?

I'm slightly confused at how this builds as the `.js` files we reference in the imports shouldn't exist!

AshCorr (approved):
Looks good! Little bit sad we have to specify `.js` extensions now, but its better than having 3 tsconfig files!

At some point we may want to look at just [re-initializing](https://github.com/guardian/cdk/blob/main/docs/setting-up-a-gucdk-project.md) CDK so that we can take advantage of some of the recent changes in CDK like the upgrade to Eslint 9, that might solve some of the file extension issues too!

PR #1320, opened 2026-07-29
Create pnpm workspace.

akinsola-guardian:
## What does this change?
This PR introduces a pnpm workspace at the repository root to manage the cdk and synthetics packages together, and attempts to centralize shared tooling dependencies into a new root package.json.
## Why?
To make it easier to manage dependencies. 

This was tested locally and in CODE.

### DIFF
--- a/cdk/package.json
+++ b/cdk/package.json
@@ -15,21 +15,21 @@
 	},
 	"devDependencies": {
 		"@guardian/cdk": "64.1.0",
-		"@guardian/eslint-config": "^16.0.0",
-		"@guardian/prettier": "11.0.0",
-		"@guardian/tsconfig": "^2.0.0",
-		"@types/jest": "^30.0.0",
-		"@types/node": "26.1.1",
+		"@guardian/eslint-config": "catalog:",
+		"@guardian/prettier": "catalog:",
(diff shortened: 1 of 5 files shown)


cdk/package.json
```diff
@@ -15,21 +15,21 @@
 	},
 	"devDependencies": {
 		"@guardian/cdk": "64.1.0",
-		"@guardian/eslint-config": "^16.0.0",
-		"@guardian/prettier": "11.0.0",
-		"@guardian/tsconfig": "^2.0.0",
-		"@types/jest": "^30.0.0",
-		"@types/node": "26.1.1",
+		"@guardian/eslint-config": "catalog:",
```
brunocatarinoguardian, line 18: Didn't know this. Interesting! Where do the version numbers come from?
Would this still work for someone who hasn't pre-downloaded all dependencies locally?

brunocatarinoguardian (approved):

cdk/package.json
akinsola-guardian: It's defined in the `pnpm-workspace.yaml` file. https://pnpm.io/catalogs

Yes, it should work. I ran make clean, then make install and it worked fine.

